PR #233, opened 2019-05-13
Add ADR 001 and ADR 002: Minimum supported Python versions

### LINKED ISSUE #167: Support latest 2 Python versions
Currently Home Assistant tracks Debian stable to decide on the minimum supported version of Python. This is currently Python 3.5 which was released on September 13, 2015!

Instead of focusing on missing features in Python 3.5.2 (a lot), I want instead to talk about changing our approach on this. Instead of following Debian Stable, I suggest that we support the latest 2 versions of Python. That will be 3.6 (released December 23, 2016) and 3.7 (released June 27, 2018). (Python 3.8 is not scheduled for release until the end of the year)

As discussed [earlier](https://github.com/home-assistant/architecture/issues/126#issuecomment-457819615), we're now aiming to make sure things work on Hass.io, Docker and Hassbian. So let's take a look at the impact:

The impact of this:
 - Hass.io: none (already 3.7)
 - Docker: none (already 3.7)
 - Hassbian: This is Debian based and so is, like Debian stable, currently supports Python 3.5.2 (I think!). Hassbian runs a script on first boot to install Home Assistant, and we can make this script install the latest Python instead. For upgrading existing installations, it is easy because the config lives in another directory and dependencies are installed inside the venv. All that is needed is:

```bash
# Assuming this is a dir with a venv with Home Assistant installed
rm -rf homeassistant
mkdir homeassistant
python3.7 -m venv homeassistant
cd homeassistant
source bin/activate
pip3 install homeassistant
```

## Implementation plan

- Hassbian implements installing the latest Python on first boot ([instructions](https://gist.github.com/SeppPenner/6a5a30ebc8f79936fa136c524417761d))
- For 4 releases (2 months), print a warning if users are running a soon to be deprecated Python version. 
- Drop Python 3.5.2 support

## Plan for new Python releases

- When a .0 release comes out (ie 3.8.0), start printing a warning for the release that will be dropped
- After at least 2 months and when a .1 release is out, drop support for the oldest version of Python.

CC @home-assistant/hassbian

balloob:
@hunterjm proposed that we start finalizing decisions once the architecture issue is done, but not yet perse implemented. ~Name inspired by Python/Django: HEP stands for Home Assistant Enhancement Proposal.~

ADR stands for Architecture Decision Record.

Fixes #167


MartinHjelmare:
I think we need docs for the HEPs. What should they include? What should they not include? Who should write them? When should they be written?

balloob:
I agree.

@hunterjm help 🙈

hunterjm:
So my proposal was more along the lines of adopting a more simple architecture decision record.  These don't need to be anywhere near as detailed as PEP documents, and if we decide to go with ADRs, I don't think we should rename them to something else.

ADRs have made it to [ThoughtWorks Radar's Adopt list](https://www.thoughtworks.com/radar/techniques/lightweight-architecture-decision-records) on many occasions, and we have adopted them in my org at work.

The records can be as simple as recording a [Status/Context/Decision/Consequences](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md), or can be more detailed by also documenting the [considered approaches](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_madr.md).

My recommendation would be to continue using issues in the architecture repository to have discussions around architecture, and possibly providing an issue template.  Once the discussion has come to a conclusion in the issue, the context and result of that discussion should be recorded in this repo with an official ADR.

Reading around ADRs:
[Documenting Architecture Decisions](https://www.thoughtworks.com/radar/techniques/lightweight-architecture-decision-records)
[MADR](https://adr.github.io/)

balloob:
Updated the PR to follow ADR, and added ADR 001: start using ADR, ADR 002: Minimum required Python version.

andrewsayre:
This may be a stupid question, but why not use the GitHub issue as the ADR #?  Yes, there will be gaps/unused numbers, but would help with traceability.

hunterjm:
@andrewsayre - You could... You can also link the GitHub issue in the ADR markdown for traceability.

andrewsayre:
@hunterjm Yes, and it is in the commits.  I'm just wondering if we also want to use it as the numbers so there isn't confusion when referencing it in the future (is it an ADR # or a Issue #?)   :)

Maybe issues go away and we just do PR submissions now instead?

hunterjm:
@andrewsayre - I was toying with that idea.  That would give us the ability to have inline comments/etc, but might also be harder to follow, and is a bigger change from how things work today.

On the other end, once discussion is done, status just changes to Approved/Rejected and the PR gets merged.  It makes it so nobody is responsible for going back and compiling the decisions.

adr/0002-minimum-supported-python-version.md
```diff
@@ -0,0 +1,27 @@
...
+
+Date: 2019-05-13
+
+## Status
+
+Accepted
+
+## Context
+
+Architecture issue: [#167](https://github.com/home-assistant/architecture/issues/167)
```
hunterjm: I'd move this under `Date:`

adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
...
+## Context
+
+We need to record the architectural decisions made on this project.
+
+## Decision
+
+We will use Architecture Decision Records, as [described by Michael Nygard](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).
+
+## Consequences
+
```
hunterjm: Another consequence:
> Introduces an additional burden on core project maintainers to diligently follow architecture discussions and record the results in this repo as an ADR.

awarecan:
We should add ADR index to README.md

balloob:
- Added it to README
- Updated ADR-001 consequence
- Added Issue template that follows ADR guidelines
- Added empty ADR file

PR #244, opened 2019-05-23
Create 0003-monitor-condition-and-data-selectors.md

pvizeli:


adr/0003-monitor-condition-and-data-selectors.md
```diff
@@ -0,0 +1,33 @@
...
+
+Architecture issue: [#100](https://github.com/home-assistant/architecture/issues/100)
+
+## Status
+
+Accepted
+
+## Context
+
+Home Assistant components use mostly `CONF_MONITORED_CONDITIONS` or other config options that allow user to select which data the interface should send to backend. The user need known what the different data points means on setup of the component/platform. Later he can select the entity again, they he want see on UI.
```
balloob:
```suggestion
A lot of Home Assistant integrations use config options like `CONF_MONITORED_CONDITIONS` to allow the user to select which data the integration should expose from the data. This means that the user needs to know what the different data points means while setting up the component/platform. While configuring it's Lovelace UI, the user has the option to include the entity or not. This means that we allow the user to pick twice.
```

adr/0003-monitor-condition-and-data-selectors.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Decision
+
+Home Assistant components/platforms should grap all available data and send this to backend. Available means data they are anyway sended from API to instance or if they need only spare resource to get all data.
```
balloob:
```suggestion
Integrations should expose all available data to the backend if that data is fetched in a single API request.
```

adr/0003-monitor-condition-and-data-selectors.md
```diff
@@ -0,0 +1,33 @@
...
+
+We should redouce selector logic mostly to end point if it make sense in context of interface. User should not read available documentations and API descrition to find out which data they want have.
```
balloob:
```suggestion
Integrations should only include selector logic if it make sense in the context of interface, like if it would require extra requests. User should not have to read the available documentation and API descriptions to find out which data they want have.
```

adr/0003-monitor-condition-and-data-selectors.md
```diff
@@ -0,0 +1,33 @@
...
+
+      Fetch            Manage           View
+  -------------      ---------      ------------
+  | Interface | ---> | Core  | ---> | Frontend |
+  -------------      ---------      ------------
+```
+
+## Consequences
+
+We don't merge PRs with useless data selectors. And remove available uneeded selectors to Home Assistant 1.0
```
MartinHjelmare:
```suggestion
We don't merge PRs with data selectors. We remove existing unneeded selectors until Home Assistant 1.0.
```

adr/0003-monitor-condition-and-data-selectors.md
```diff
@@ -0,0 +1,33 @@
...
+We focus us to a good view/frontend to select and group available data.
```
MartinHjelmare:
```suggestion
We focus on a good view/frontend to select and group available data.
```

adr/0003-monitor-condition-and-data-selectors.md
```diff
@@ -0,0 +1,33 @@
...
+
+Architecture issue: [#100](https://github.com/home-assistant/architecture/issues/100)
+
+## Status
+
+Accepted
+
+## Context
+
+A lot of Home Assistant integrations use config options like `CONF_MONITORED_CONDITIONS` to allow the user to select which data the integration should expose from the data. This means that the user needs to know what the different data points means while setting up the component/platform. While configuring it's Lovelace UI, the user has the option to include the entity or not. This means that we allow the user to pick twice.
```
MartinHjelmare:
```suggestion
A lot of Home Assistant integrations use config options like `CONF_MONITORED_CONDITIONS` to allow the user to select which data the integration should expose from the data. This means that the user needs to know what the different data points mean while setting up the integration. While configuring its Lovelace UI, the user has the option to include the entity or not. This means that we allow the user to pick twice.
```

Issue #249, opened 2019-06-06
Use Black for Code Formatting


bachya:
## Context

A long time ago, code formatting via `yapf` was discussed here: https://github.com/home-assistant/home-assistant/issues/9793. I couldn't find such a discussion in this repo, so resurfacing it as a holistic architectural issue for discussion.

**EDIT:** @balloob kindly reminded me that it was discussed here, as well: https://github.com/home-assistant/home-assistant/issues/16634

## Proposal

I propose that we make [Black](https://github.com/python/black) the code style and formatter of choice for the HASS back-end. After using `autopep8` and `yapf`, I like that Black handles a lot by itself:

>Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.
>
>Blackened code looks the same regardless of the project you're reading. Formatting becomes transparent after a while and you can focus on the content instead.
>
>Black makes code review faster by producing the smallest diffs possible.

In the 206 PRs I've merged into the HASS back-end, I would estimate that 70-80% had styling issues that either (a) failed a PR check or (b) invited additional PR commentary that, although valid, is a waste of the reviewer's and my time. This has forced me to spend countless hours fiddling with linters (and altering zillions of options), tweaking editor configs, plodding through fixes, and otherwise not spending time adding meaningful functionality to HASS.

I recently updated all of my PyPi libraries to use Black and love how it takes formatting control out of my hands, produces beautiful-looking code, and frees my mind up to focus on functionality. I've even paired it with [`pre-commit`](https://pre-commit.com/) so that any PRs submitted to those libraries are automatically styled by Black upon commit.

I have to believe that the proper application of Black would greatly reduce the amount of time we spend on reviewing style/formatting.

## Consequences

Of all the Python formatters, Black provides the fewest configuration options (requiring that control over "sane formatting" be ceded to it), which may run contrary to some people's preferences. Additionally, it is technically a beta product; although the group assures the community that it will focus primarily on bug fixes, there is no guarantee future updates won't hurt something.

Although it's possible that running Black on the entire codebase at once will produce proper results, it's likely safer to stage such an effort; thus, there is a "consequence" of this taking decent time to implement.

balloob:
You missed this one ;-) https://github.com/home-assistant/home-assistant/issues/16634

Thanks for opening it as an architecture issue.

[...]

PR #257, opened 2019-07-07
Add ADR-0005 for code formatting via Black

bachya:
Per @pvizeli's request, this creates an ADR for https://github.com/home-assistant/architecture/issues/249.

### DIFF
--- a/adr/0005-code-formatting.md
+++ b/adr/0005-code-formatting.md
@@ -22,10 +22,10 @@ base. In place, a codebase with such formatting is:
 - We use [Black](https://github.com/python/black) as code formatter for the back-end codebase.
 - We use Black's default options.
 - We utilize a bot (such as [the black_out bot](https://github.com/Mariatta/black_out)) to ensure PRs are appropriately formatted before being accepted.
-- We recommend and document the set-up of Black, `pylint`, and `flake8` pre-commit hooks in the development environment.
+- We recommend and document the set-up of `Black`, `pylint`, and `flake8` in the development environment.
 
 ## Consequences
 
-- We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503` to `setup.cfg`.
-- We retain pylint, but disable all formatting checks.
+- We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503, E203` to `setup.cfg`.
+- We retain pylint, but disable `bad-continuation`.
 - Initially, we create a single PR to Black-ify the entire codebase (`homeassistant`) and tests (`tests/homeassistant`).
--- a/adr/0005-code-formatting.md
+++ b/adr/0005-code-formatting.md
@@ -28,4 +28,3 @@ base. In place, a codebase with such formatting is:
 
 - We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503, E203` to `setup.cfg`.
 - We retain pylint, but disable `bad-continuation`.
-- Initially, we create a single PR to Black-ify the entire codebase (`homeassistant`) and tests (`tests/homeassistant`).
--- a/adr/0005-code-formatting.md
+++ b/adr/0005-code-formatting.md
@@ -1,4 +1,4 @@
-# 4. Code Formatting
+# 5. Code Formatting
 
 Date: 2019-07-07
 
@@ -25,4 +25,5 @@ base. In place, a codebase with such formatting is:
 
 ## Consequences
 
-- We will need to determine what to do with style checking by `pylint`, `flake8`, etc.
+- Flake8: we set `max-line-length = 88` and `ignore = E501, W503` to `setup.cfg`.
+- Pylint: we disable `bad-continuation` on `pylintrc`
--- a/adr/0005-code-formatting.md
+++ b/adr/0005-code-formatting.md
@@ -21,10 +21,10 @@ base. In place, a codebase with such formatting is:
 
 - We use [Black](https://github.com/python/black) as code formatter for the back-end codebase.
 - We use Black's default options.
-- We utilize a bot (such as [the black_out bot](https://github.com/Mariatta/black_out)) to ensure PRs are appropriately formatted before being accepted.
-- We recommend and document the set-up of `Black`, `pylint`, and `flake8` in the development environment.
+- We CI to ensure PRs are appropriately formatted before being accepted.
+- We recommend and document the setup of `Black`, `pylint`, and `flake8` in the development environment.
 
 ## Consequences
 
 - We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503, E203` to `setup.cfg`.
-- We retain pylint, but disable `bad-continuation`.
+- We retain pylint, but disable all formatting checks.
--- a/adr/0005-code-formatting.md
+++ b/adr/0005-code-formatting.md
@@ -27,3 +27,5 @@ base. In place, a codebase with such formatting is:
 
 - Flake8: we set `max-line-length = 88` and `ignore = E501, W503` to `setup.cfg`.
 - Pylint: we disable `bad-continuation` on `pylintrc`
+
+On initial step, we create one big PR they change the format again codebase (tests/homeassistant).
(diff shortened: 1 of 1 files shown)


adr/0005-code-formatting.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Proposal
+
+- We use [Black](https://github.com/python/black) as the de-facto code formatter for the entire back-end codebase
+- We use Black's default options
+- We utilize a bot (such as [the black_out bot](https://github.com/Mariatta/black_out)) to ensure PRs are appropriately formatted before being accepted
+
+## Consequences
+
+- We will need to determine what to do with style checking by `pylint`, `flake8`, etc.
```
MartinHjelmare: We should determine this before finalizing this proposal. Determining it shouldn't be part of consequences. Rather what changes to pylint and flake8 config could be part of consequences.

bachya: What's the process for determining that? So far, there are two proposals:

* Mine: https://github.com/home-assistant/architecture/issues/249#issuecomment-499709271
* @frenck's: https://github.com/home-assistant/architecture/issues/249#issuecomment-499803823

MartinHjelmare: Make a suggestion on all rules to disable for pylint and flake8 etc. Then we come to a decision.

adr/0005-code-formatting.md
```diff
@@ -0,0 +1,28 @@
...
+Proper code formatting is vital to the long-term viability and maintainability of a code
+base. In place, a codebase with such formatting is:
+
+- easier to review
+- easier to diff against previous commits
+- more likely to be free of linting errors
+
+## Proposal
+
+- We use [Black](https://github.com/python/black) as the de-facto code formatter for the entire back-end codebase
```
ties:
```suggestion
- We use [Black](https://github.com/python/black) as code formatter for the back-end codebase
```

To me, de-facto and entire are weasel words and I do not think they are needed here

adr/0005-code-formatting.md
```diff
@@ -0,0 +1,28 @@
...
+- We use Black's default options
+- We utilize a bot (such as [the black_out bot](https://github.com/Mariatta/black_out)) to ensure PRs are appropriately formatted before being accepted
+
```
ties:
```suggestion
- We recommend and document the set-up of Black, `pylint`, and `flake8` pre-commit hooks in the development environment

```

adr/0005-code-formatting.md
balloob: I'm with Frenck, let's disable all formatting checks because we are no longer manually formatting things.

adr/0005-code-formatting.md
```diff
@@ -0,0 +1,31 @@
...
+- easier to review
+- easier to diff against previous commits
+- more likely to be free of linting errors
+
+## Proposal
+
+- We use [Black](https://github.com/python/black) as code formatter for the back-end codebase.
+- We use Black's default options.
+- We utilize a bot (such as [the black_out bot](https://github.com/Mariatta/black_out)) to ensure PRs are appropriately formatted before being accepted.
+- We recommend and document the set-up of Black, `pylint`, and `flake8` pre-commit hooks in the development environment.
```
rytilahti: Are not `pylint` and `flake8` precommit hooks offtopic for the purpose of this ADR? Not that I think that they shouldn't be used, but in my view they are not relevant to the topic at hand (outside the configuration changes described below).

adr/0005-code-formatting.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Consequences
+
+- We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503` to `setup.cfg`.
```
rytilahti: E203 should also be added to the ignore list for flake8, per https://github.com/python/black#slices

adr/0005-code-formatting.md
```diff
@@ -0,0 +1,31 @@
...
+- We retain pylint, but disable all formatting checks.
```
MartinHjelmare: Have we reviewed what rules will be disabled by disabling all pylint formatting?

adr/0005-code-formatting.md
ties: Kind of.

You want code to be consistent and valid at every commit. Since this ADR has become broader than adding `black` itself (since it changes `flake8` and `pylint` settings), I thought these were relevant.

In my opinion the minimal recommendation, while being on topic, of only adding black commit hooks definitely is wrong.

I would be in favor of opening up the scope of the ADR to concern all code formatting and linting. I think that the concerns about `flake8` and `pylint` show that a slightly more holistic approach may be needed here.

frenck: I'm in favor of adding `pre-commit` to our project, nevertheless, this should really be a separate ADR.

ties: Enforced coding standards _are_ already a part of the project. Both through CircleCI and the recommendation to set-up flake8 pre-commit hooks in the [development and testing](https://developers.home-assistant.io/docs/en/development_testing.html#preventing-linter-errors) instructions. I see adding removing flake8 rules and adding black without pre-commit as a regression compared to the status quo.

Potentially there is a trivial ADR on coding standards here.

Specific to this ADR, I do think you want black to be ran locally? Iteratively fixing issues after CircleCI runs fails or issue comments is noisy and imo is an anti-pattern.

frenck: Let discuss this in the ADR issue instead of the PR.

adr/0005-code-formatting.md
```diff
@@ -22,10 +22,10 @@ base. In place, a codebase with such formatting is:
...
 - We utilize a bot (such as [the black_out bot](https://github.com/Mariatta/black_out)) to ensure PRs are appropriately formatted before being accepted.
-- We recommend and document the set-up of Black, `pylint`, and `flake8` pre-commit hooks in the development environment.
+- We recommend and document the set-up of `Black`, `pylint`, and `flake8` in the development environment.
 
 ## Consequences
 
-- We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503` to `setup.cfg`.
-- We retain pylint, but disable all formatting checks.
+- We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503, E203` to `setup.cfg`.
+- We retain pylint, but disable `bad-continuation`.
```
frenck: We should disable all formatting check with pylint, instead, check those with Black itself.
To captains on a ship is a bad idea when it comes to formatting.

MartinHjelmare: Have you checked what rules will be disabled?

frenck: Yes, I did.

https://github.com/PyCQA/pylint/blob/master/pylint/checkers/format.py#L127

MartinHjelmare: Thanks! Looks good.

adr/0005-code-formatting.md
```diff
@@ -0,0 +1,31 @@
...
+- We use [Black](https://github.com/python/black) as code formatter for the back-end codebase.
+- We use Black's default options.
+- We utilize a bot (such as [the black_out bot](https://github.com/Mariatta/black_out)) to ensure PRs are appropriately formatted before being accepted.
+- We recommend and document the set-up of `Black`, `pylint`, and `flake8` in the development environment.
+
+## Consequences
+
+- We retain Flake8, but set `max-line-length = 88` and `ignore = E501, W503, E203` to `setup.cfg`.
+- We retain pylint, but disable `bad-continuation`.
+- Initially, we create a single PR to Black-ify the entire codebase (`homeassistant`) and tests (`tests/homeassistant`).
```
balloob: I don't think that we need to further specify what `codebase` entails.

adr/0005-code-formatting.md
pvizeli: I run black with these settings on Hass.io and there is no problem or reason to disable all formattings

balloob: I would say that the reason is too many cooks in the kitchen. Since the user is unable to control formatting, we shouldn't lint for it either.

PR #289, opened 2019-09-17
Require config yaml and flow

### LINKED ISSUE #142: Config structure for embedded platforms
It's nice that we streamline the ways of introducing new component and/or platform modules to home assistant. Now, we know that there should always be a new component package and a platform should go under that component.

But what about config structure? Should we keep the platforms (itach) under a domain (remote) config structure also for new platforms? Or should we ask to set up a component that has the config?

_Originally posted by @MartinHjelmare in https://github.com/home-assistant/home-assistant/pull/20809#issuecomment-461298673_

### LINKED ISSUE #143: Strategy for config methods
With the 0.87.0 release, [the top comment](https://www.reddit.com/r/homeassistant/comments/ao4jhv/087_smartthings_areas_and_entity_registry_ui/efyb7t9) on [the Reddit post](https://www.reddit.com/r/homeassistant/comments/ao4jhv/087_smartthings_areas_and_entity_registry_ui/) says:

>These kind of statements make me sad:

>>The SmartThings component is configured exclusively through the front-end. Manual setup through configuration.yaml is not available at this time.

Subsequent replies to that comment:

>The configuration is getting more and more fragmented. Some via config, some via ui. Then you get to the front-end where you have to make a hard choice. Config or front-end, not both, not even a little.

>I like being able to manually set/see everything. This is why I use yaml mode for Lovelace still. I'm sure people who use Unix on a regular basis are just more comfortable with configuration files rather than a black box UI

>I'm having a small pity party about not being able to do everything in my IDE anymore, but I'll get over it.

>To add onto what -Smokin- was saying, easier troubleshooting and deployment. What happens when HA is offline and I want to make a change? What if something caused by the SmartThings integration is causing an issue? Might not be the case now but I think that the ability to manage everything via terminal/config is the best approach. And, this has no impact who prefer the UI interface.

>Removing manual control via configurations files from poweruser to make it easier for 'non-tech' people is never a good idea. Remember that most of us use HA because we don't support cloud control over our things, we like to control our things. Saying that, I really like to have my configuration on configuration files so I can change them to exactly I want or backup the files, share it, etc.

>I don't mean to harp on the UI-only configs yet again, but the .storage directory has seriously complicated my configuration and usability within HA.
>
>I know there's some great features coming out of them but recently I've had to restart several times and manually edit registries because I wanted to rename something and there were conflicts. I appreciate that these help UI users but it's made things harder for config-based, advanced users.
>
>Edit: also, it's not clear to me as a user whether that hidden directory should be backed up or committed to source control.

>Having configuration as a file means you can use a bunch of tools - rsync, git versioning, diff, grep, sed, bulk rename, etc - that you simply cannot if it is stored in a database that's only meant to be accessed through the UI. Those command line tools can make long-term maintenance easier: you can check your git log for things you've added, do find-and-replace operations on your configuration, archive old setups as you retire devices or take them off line, etc. And if you run into issues and need to troubleshoot or just decide to get rid of configuration clutter, you can just create a backup/git version, blank your config files, and then use diff/merge techniques to pick and choose what you want; with the database approach, you have to start from scratch without carrying anything over.
>
>A few examples I can think of: HASS recently had a change that made it so entity names would no longer end up having double underscores - this affected a bunch of my entity names, and fortunately all it took was one sed command to fix it in the YAML, instead of having to edit 100+ entity names in the UI. Or when I switched to a new NAS and archived the configuration to integrate with the old one - I recently found a use for my old NAS, so all I had to do was dig the file up out of my archive and add one #include statement, instead of having to reconfigure the dozen or so entities and automations from scratch through the UI.
>
>Perhaps a good intermediary would be to make it so the configuration file is the permanent storage, and the database are more ephemeral changes on top of that? So by default, the configuration is read from the YAML file; when you are ready to make changes through the UI, you turn on "Edit Mode", and all changes are applied to the database, but not to the file; and then when you leave "Edit Mode" you have the option to commit the changes back to the file, or discard the changes and reload the file? In a worst-case situation, you can nuke the database and only lose your recent changes.
>
>That said, I do want to point out that this is nitpicking for the sake of providing you feedback - the UI approach might not be my preferred option, but it works just fine.

>Having it in the config file allows me to track changes via git and also share configurations with others. Keeping it in the UI only makes that much more difficult.

>I agree with this sentiment. Also, I just prefer yaml because I’m comfortable with it and I like having a central place that I can edit my config in a terminal or editor.

>Sharing configurations will now be sets of screenshots and not code blocks.

>Disclaimer: I'm a linux admin by day, not a typical user.
>
>Because text configs are much much easier to manage than DB based configs. Text files are repeatable and easily manageable.
>
>Text files also allow me to use config management so I can have repeatable builds so it's trivial for me to change a variable and have a full clone of my hass setup in a dev machine.
>
>Note: When I say text files I mean files that I create and modify only, not the program writing config out to a text file.
>
>When everything is configured purely with the UI this generally means the config is stored in a database of some kind (or shudders written to a text file) this means I now need to worry about making, verifying, and testing database backups.
>
>Perhaps it has already been discussed elsewhere, but this indicates that (a) the config UI is _not_ the preferred user experience for a subset of the user base and (b) there should be a well-defined config strategy going forward. Possible options below.

# 1. Promote a Hybrid Model for All Integrations

In this scenario, all integrations would support configuration in both a UI config entry and via `configuration.yaml` – they cannot cherry-pick the one they prefer. Config entries would focus on the "quickest possible route" (i.e., the lowest number of configuration options to get the integration up and running) and `configuration.yaml` would focus on exposing/tweaking all options.

# 2. Promote UI Config Entries as "The Future"

In this scenario, all integrations would be slowly migrated away from `configuration.yaml` and interacted with via UI config entry only. Over time, the config UI would be expanded to allow more configuration options for each integration.

# 3. To Each Their Own

In this scenario, each integration developer is left to their own devices. If they desire one method or the other (or both), implementation is up to their discretion.

MartinHjelmare:
Closes #142, closes #143 

Here's a related discussion:
https://github.com/home-assistant/home-assistant/pull/26648#issuecomment-531514841

### DIFF
--- a/adr/0007-require-config-yaml-and-flow.md
+++ b/adr/0007-require-config-yaml-and-flow.md
@@ -24,7 +24,7 @@ These many options for configuration also impacts how the integration is structu
 
 1. Integrations that only have a single platform puts the configuration under a platform key and just uses `async_setup_platform` to set up the platform.
 2. Integrations that have many platforms sometimes centralize the configuration to the integration domain name key and loads platforms via `discovery.async_load_platform`, which in turn calls `async_setup_platform`.
-3. If the integration has multiple platforms but doesn't have a centralized config, the user needs to enter config under the integration domain to set up the integration and each platform key to set up each platform. This loads each platform via `setup_platform`.
+3. If the integration has multiple platforms but doesn't have a centralized config, the user needs to enter config under the integration domain to set up the integration and each platform key to set up each platform. This loads each platform via `async_setup_platform`.
 4. Integrations that support config flow, uses `async_setup_entry` to set up the config entry for the integration and `config_entries.async_forward_entry_setup` to set up platforms. Setting up platforms via `discovery.load_platform` is not available for these integrations.
 
 All backend APIs in point 1-3 above also have sync versions, to further increase the options.
@@ -45,8 +45,8 @@ On more than one occassion Home Assistant official representatives have said tha
 ## Proposal
 
 1. We limit the configuration structure to one way for new integrations:
-  1. We require all yaml config for an integration to be located under the integration domain key in config yaml, for all new integrations.
-  2. We require both config yaml and config flow support for all new integrations.
+  - We require all yaml config for an integration to be located under the integration domain key in config yaml, for all new integrations.
+  - We require both config yaml and config flow support for all new integrations.
 2. We do not allow removing support for either config yaml or config flow in existing integrations.
 
 
--- a/adr/0007-require-config-yaml-and-flow.md
+++ b/adr/0007-require-config-yaml-and-flow.md
@@ -1,0 +1,64 @@
+# 7. Require config yaml and flow
+
+Date: 2019-09-17
+
+## Status
+
+Proposed
+
+---
+
+## Context
+
+This ADR has two faces that it wants to address:
+
+1. The configuration structure disparity, and its consequences.
+2. What Home Assistant is promising users in terms of ways of configuration.
+
+
+### Config structure disparity
+
+There are currently many different ways of structuring an integration and its config in Home Assistant. We allow config yaml, either under the integration key, or under platform keys with the integration name. We allow a config flow with configuration entered via the GUI, stored in config entries. We allow importing config yaml to a config entry via config flow.
+
+These many options for configuration also impacts how the integration is structured and what Home Assistant backend APIs are used.
+
+1. Integrations that only have a single platform puts the configuration under a platform key and just uses `async_setup_platform` to set up the platform.
+2. Integrations that have many platforms sometimes centralize the configuration to the integration domain name key and loads platforms via `discovery.async_load_platform`, which in turn calls `async_setup_platform`.
+3. If the integration has multiple platforms but doesn't have a centralized config, the user needs to enter config under the integration domain to set up the integration and each platform key to set up each platform. This loads each platform via `setup_platform`.
+4. Integrations that support config flow, uses `async_setup_entry` to set up the config entry for the integration and `config_entries.async_forward_entry_setup` to set up platforms. Setting up platforms via `discovery.load_platform` is not available for these integrations.
+
+All backend APIs in point 1-3 above also have sync versions, to further increase the options.
+
+This multitude of options for config and integration structure makes it:
+
+1. Harder for users to know how to configure the integrations.
+2. Harder for contributors to know what way to implement and support.
+3. Harder for Home Assistant to support confused users and contributors and harder to support and extend a disparate code base.
+
+
+### Home Assistant's promise to users
+
+On more than one occassion Home Assistant official representatives have said that config yaml will not go away as a way of configuring Home Assistant, even after introducing config flow. If we are serious about that promise, Home Assistant needs to take responsibility for the way that integrations offer ways of configuration. If we do not require integrations to offer yaml as a way of configuring, we have to change what we say to users and inform them that we can not promise that yaml will stay.
+
+---
+
+## Proposal
+
+1. We limit the configuration structure to one way for new integrations:
+  1. We require all yaml config for an integration to be located under the integration domain key in config yaml, for all new integrations.
+  2. We require both config yaml and config flow support for all new integrations.
+2. We do not allow removing support for either config yaml or config flow in existing integrations.
+
+
+This proposal will solve both faces of the ADR purpose.
+
+---
+
+## Consequences
+
+1. The contributor work of implementing a single platform will become greater. Config flow requires unit tests, which not all contributors are familiar with.
+2. New integration PRs will become longer and require more from code review.
+3. Integration maintenance by developers will increase since they have to support both config yaml and config flow.
+4. It will be easier to recommend approaches for implementation to contributors and we can simplify our development documentation, with fewer approved ways of structuring an integration.
+5. We will have fewer breaking changes, since integrations will implement the best way from the beginning with a centralized config and config entries support migration.
+6. We can send a clearer message to users about how configuration is done in Home Assistant.


MartinHjelmare:
CC @balloob and @Kane610

Jc2k:
Are there exceptions to this? For example, some integrations have really good user experiences with config flow. Take my homekit_controller stuff - automatic discovery with zeroconf, enter the pin code, crypto keys are exchanged and stored. And I think Hue is similar with the "press button now" UI?

I am lucky as an existing integration maintainer because i'm "grandfathered in", but what about new integrations where there are rich UI based pairing experiences similar to these?

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
+# 7. Require config yaml and flow
```
frenck:
```suggestion
# 7. Require config YAML and flow
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+
+Proposed
+
+---
+
+## Context
+
+This ADR has two faces that it wants to address:
+
+1. The configuration structure disparity, and its consequences.
```
frenck:
```suggestion
1. The configuration structure disparity and its consequences.
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+2. What Home Assistant is promising users in terms of ways of configuration.
+
+
+### Config structure disparity
+
+There are currently many different ways of structuring an integration and its config in Home Assistant. We allow config yaml, either under the integration key, or under platform keys with the integration name. We allow a config flow with configuration entered via the GUI, stored in config entries. We allow importing config yaml to a config entry via config flow.
```
frenck:
```suggestion
There are currently many different ways of structuring an integration and its config in Home Assistant. We allow config YAML, either under the integration key or under platform keys with the integration name. We allow a config flow with configuration entered via the GUI, stored in config entries. We allow importing config YAML to a config entry via config flow.
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+
+These many options for configuration also impacts how the integration is structured and what Home Assistant backend APIs are used.
```
frenck:
```suggestion
These many options for configuration also impact how the integration is structured and what Home Assistant backend APIs are used.
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+This multitude of options for config and integration structure makes it:
+
+1. Harder for users to know how to configure the integrations.
+2. Harder for contributors to know what way to implement and support.
+3. Harder for Home Assistant to support confused users and contributors and harder to support and extend a disparate code base.
+
+
+### Home Assistant's promise to users
+
+On more than one occassion Home Assistant official representatives have said that config yaml will not go away as a way of configuring Home Assistant, even after introducing config flow. If we are serious about that promise, Home Assistant needs to take responsibility for the way that integrations offer ways of configuration. If we do not require integrations to offer yaml as a way of configuring, we have to change what we say to users and inform them that we can not promise that yaml will stay.
```
frenck:
```suggestion
On more than one occasion Home Assistant official representatives have said that config YAML will not go away as a way of configuring Home Assistant, even after introducing config flow. If we are serious about that promise, Home Assistant needs to take responsibility for the way that integrations offer ways of configuration. If we do not require integrations to offer YAML as a way of configuring, we have to change what we say to users and inform them that we can not promise that YAML will stay.
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+
+---
+
+## Proposal
+
+1. We limit the configuration structure to one way for new integrations:
+  - We require all yaml config for an integration to be located under the integration domain key in config yaml, for all new integrations.
```
frenck:
```suggestion
  - We require all YAML config for an integration to be located under the integration domain key in config YAML, for all new integrations.
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+  - We require both config yaml and config flow support for all new integrations.
```
frenck:
```suggestion
  - We require both config YAML and config flow support for all new integrations.
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+2. We do not allow removing support for either config yaml or config flow in existing integrations.
```
frenck:
```suggestion
2. We do not allow removing support for either config YAML or config flow in existing integrations.
```

adr/0007-require-config-yaml-and-flow.md
```diff
@@ -0,0 +1,64 @@
...
+
+This proposal will solve both faces of the ADR purpose.
+
+---
+
+## Consequences
+
+1. The contributor work of implementing a single platform will become greater. Config flow requires unit tests, which not all contributors are familiar with.
+2. New integration PRs will become longer and require more from code review.
+3. Integration maintenance by developers will increase since they have to support both config yaml and config flow.
```
frenck:
```suggestion
3. Integration maintenance by developers will increase since they have to support both config YAML and config flow.
```

frenck:
I agree, partly. I think we should hold up to that promise, however, I've already noticed a couple of users in distress on not getting the difference, which I think need to be addressed.

- A user that is doing both, config YAML & flow and getting lost in the process.
- If a user did both, it is not clear which configuration is used. We do strive for YAML overriding flow, however, that is not always the case, nor is it visible.
- No clear path of migration from YAML <-> Flow for the user.

For the above reasons alone, I did not add YAML support to the integrations I provided.

A suggestion, which is kinda out of scope for this ADR maybe, it that we should think about handling the flow (e.g., by disabling it) or some other visible indicator to show the user the settings are from YAML/

cgarwood:
Here's my 2 cents...

IMO the only reason to keep configuration.yaml around for things like integration/platform setup was to overcome the fact that you couldn't update a config entry's settings without removing it and setting it back up. In the case of zwave, that meant that if your usb path changed you either had to remove the zwave integration and re-set it up (thus getting a new `network_key` as well) or edit the usb_path in the `.storage/core.config_entries` file. So I added support in the zwave component so that  a usb_path in configuration.yaml would override the config entry.

However, I'd love to see that _go away_ and just be configured from the UI.

The big argument from the community for keeping YAML around is because people share configs and post it on GitHub. That's great, _but _  almost all the integration setup blocks in configuration.yaml are just specifying IP addresses/usernames/passwords in order to make the connection. There's zero value in that being shared on GitHub and nobody looking at configs on GitHub is looking for how to specify the connection details for `deconz:`, they're looking for things like scripts, automations, templates, etc (which are still YAML, and I've heard no talk of changing that).

The next big thing I hear is "but what about backups?" - back up your `.storage` folder along with everything else and you're set 🤷‍♂ 

"But I have a system set up where i commit config changes to `git` and they get linted and CI uploads them to my HA server and restarts it" - Cool, keep that for your automations, but use the _much simpler_ UI config to set up connection information for an integration.


I see no reason to force integrations to maintain YAML config, especially now that we have config entry options and the settings in the config entry can be updated from the UI. It adds complexity to the component (especially since it's up to the component how it imports or merges those configs), is confusing for the users, and requires extra documentation.

dmulcahey:
I agree with @cgarwood and @Jc2k. I think the config flows are much more user friendly. I would think that approachability for the common non-developer user should be the focus in the future no? The UI driven config is much more natural for users IMHO. How many times have we had to help users who have been struggling with configuration because tabbing / spacing is off??? This isn’t a friendly experience for new users (especially non developers / non techies)  and I would think that we would want to navigate away from this. 

All of that said... couldn’t we build some core helper (config helper) / utility that automatically provides YAML support off of the config flow code for each integration? This way the integration devs build 1 but the platform provides both? This could potentially solve the ordering / precedence issue too. The api would allow you to query for configuration and it could handle checking YAML then the config entry)... just a thought.

MartinHjelmare:
> Why requiring config YAML and not the config flow?

Maybe this wasn't clear, but this ADR does require all new integrations to support both config yaml and config flow.

MartinHjelmare:
Since this is a hot topic, I'll wait with changing any text until the discussion settles down.

Kane610:
I really don't like the aim for this ADR.

Just because configuration.yaml isn't going away isn't the same as saying an integration can't be config flow only. It is not a motivation to force it on all integrations.

Having this as a requirement will really stifle progress going forward. It would mean all new integrations need to start at least with configuration.yaml support and at a different stage the config entry support since previous recommendations have been to keep PRs as small as possible. If anything it will probably lead to slower convergence of using config flow and in the current state of things make less use of device registry and area registry.

Personally I'm moving away from configuration.yaml on my integrations, I don't see a reason to support both and I much prefer config flow over configuration.yaml.

MartinHjelmare:
To clarify, the aim of this ADR is this:
> This ADR has two faces that it wants to address:

> The configuration structure disparity, and its consequences.
What Home Assistant is promising users in terms of ways of configuration.

That aim and how we solve that is not the same thing. The current text reflects my proposal to reach that aim. If we come to a different solution that reaches the aim, the ADR will still be useful.

balloob:
I agree with all of you 😆 

As this ADR has two parts, let me respond to both and I'll respond to some other comments along the way.

> The configuration structure disparity, and its consequences.

It's very good to make a decision and write it down how we want things to be configured moving forward. 

There are 4 different ways of doing YAML (1 top level with N configs, N top level with 1 config, 1 platform config with N configs, N platform configs with 1 config). By limiting future contributed YAML configurations to a single one, we will create a more predictable experience for YAML-based users.

If an integration supports both YAML and config entries, they should take one of either approaches:
 - It imports/syncs the YAML data to a config entry so that there is a single code path from there on (example:  `hue`).
 - If the integration has no external dependencies (devices/services), it treats YAML as an immutable data store and uses a storage backend to support a mutable data store for a UI (example: `person`, `lovelace`)

I believe that YAML's true value lies in making it easy to version control and share your creations. This means YAML is great for integrations that work with Home Assistant data: Automations, Scripts, Scenes, Lovelace.

> What Home Assistant is promising users in terms of ways of configuration.

Home Assistant is built up very modular. A lot of things are optional to implement. This makes it very easy to get an initial piece contributed. You don't have to implement unique ID, device info, etc. I don't think that having both YAML and config flow should be a requirement.

Integrations are allowed to have both, but it is up to the developer to decide how they want to spend their time. If other people want to contribute the missing configuration method (either YAML or config flow), they are free to do so.

MartinHjelmare:
If we don't require either only config yaml, only config flow or both config yaml and config flow, the disparity will continue and we won't reach the goal of this ADR.

Going back to only config yaml isn't an option. So that leaves two ways of reaching the goal, as I see it.

> If an integration supports both YAML and config entries, they should take one of either approaches:

>  -  It imports/syncs the YAML data to a config entry so that there is a single code path from there on (example: hue).
  -  If the integration has no external dependencies (devices/services), it treats YAML as an immutable data store and uses a storage backend to support a mutable data store for a UI (example: person, lovelace)

This is what we do today already. Just continuing that won't improve the situation much.

It's good to streamline the config yaml structure, but only doing that won't significantly decrease the ways we structure integrations and the backend APIs we recommend.

MartinHjelmare:
There's a third option that would allow us to decrease the disparity in the backend APIs usage: Require config flow and leave config yaml optional.

This will mean we always use the config entries API for new integrations.

MartinHjelmare:
> Integrations are allowed to have both, but it is up to the developer to decide how they want to spend their time. If other people want to contribute the missing configuration method (either YAML or config flow), they are free to do so.

What should we do if one contributor wants to remove support for config yaml, does that, and then later another contributor comes and wants to add support back for config yaml?

balloob:
I don't think that we should allow removing config yaml support. That would be an unnecessary breaking change. 

I personally don't mind the contribution start with either config.yaml or config entries. Some people are more passionate about writing a driver/Python lib, others about the integration with HA.

I think that if we have to choose between a higher bar for contributions and a less predictable way how something would be configured, I would lean to less predictable.

For what it's worth, we could explore creating more config flow helpers. We have one based on discovery which is used a lot.

pvizeli:
I agree with your statement above @balloob but with the second comment:
> I don't think that we should allow removing config yaml support. That would be an unnecessary breaking change.

I have my issue. It's hard to maintain the breaking change faced YAML and the variable/dynamic config flow in one place. So I can understand the people they remove this hard in maintaining bridge on integrations. We should not add stones on that if developers want to use the new config flow (what we want) and don't want to spend on the complexity to merge the old config with stuff they can change over hass.io/UI/... in future.

SeanPM5:
Me personally, I like the idea of requiring config flow and making yaml optional. Then there would be clear, concise, and most importantly _consistent_ instructions:
>Go to the Integrations page and press the add button where you'll be guided through our easy setup process.

Currently the user experience is kind of all over the place:
* Go to the Integrations page, sometimes, maybe? If you don't see it there and come up empty-handed, you gotta do it the other way I guess.
* Even though these two integrations were added in the _same exact version of Home Assistant_, they require being set up in two radically different ways. 
* Sometimes integrations give you a choice between config flow or yaml, but often the config flow version lacks most of the available options that you can configure in YAML. So it's not _really_ a true choice.
* Straight up the vast majority of people are not willing to learn a markup language in order to install an app. 
* Even if someone does jump through all those hoops they still have the tedious aspect of visiting the website and searching through the integrations to find the documentation, referencing a long `CONFIGURATION VARIABLES` table and figuring everything out. And then possibly dealing with duplicate keys, bad indentation and numerous other ways things can go wrong.

I think that process is far too convoluted. You wouldn't ever be able to put YAML integrations in front of your wife or parents and have them figure it out. It doesn't pass the "normal person" test. They'll say "screw this" within two minutes and walk away. 

I get wanting to make things easier for contributors. But for every contributor there's probably thousands of regular users that are struggling or giving up completely due to this aspect being too difficult.  By not putting this extra burden on contributors it's instead making things harder than they need to be for the entire userbase.

Keep in mind that users can't even really begin writing automations or doing much of anything until they get over the initial hump of adding their integrations and devices.. A lot of things are getting easy now with Device automations and Lovelace Configure UI etc but the user can't even make it to that point without adding their devices first. 

So this is a fairly crucial area to "get right" and make as easy as possible imo.

balloob:
In that case we should make it dead easy to scaffold an integration with a config flow, including tests. I am going to see if I can cook up a generator.

Adminiuga:
> I believe that YAML's true value lies in making it easy to version control and share your creations. 

Agreed, but in case when a configuration is just a "serial port path" then this value is diminished. (Side note:  maybe allow config flows be exported/imported?)

> If an integration supports both YAML and config entries, they should take one of either approaches:
> - It imports/syncs the YAML data to a config entry so that there is a single code path from there on (example: hue).

The problem with that approach: It is done only once and it is a common pitfall at least with ZHA when users update `configuration.yaml` but since it was already imported before, the new `configuration.yaml` doesn't have any effect. Very same problem mentioned by @cgarwood:
>  the fact that you couldn't update a config entry's settings without removing it and setting it back up. In the case of zwave, that meant that if your usb path changed you either had to remove the zwave integration and re-set it up (thus getting a new network_key as well) or edit the usb_path in the .storage/core.config_entries file. 

so are there ways to update an existing entry without removing the integration completely? Cause in some cases that is not feasible as it removes all custom entity ids.

balloob:
I guess the source=import that we have been using will indeed update it after the current entry has tried to set up, which is not ideal. That's a problem we could fix though.

I have just created a scaffolding script that should make it _super_ easy to create a new integration with a config flow. It comes with tests includes. https://github.com/home-assistant/home-assistant/pull/26777. With that in place, I think that we could consider making a config flow approach.

rytilahti:
I have discussed about my opinion about this (and even done some wireframes for config forms) on discord previously, so here's my $0.02: why not simply use the given configuration constraints the developers already have to define (in voluptuous) to create a Configuration object to be used by an integration?

How it would work:
1. Integration developers _MUST_ define a schema for the configuration, if any. (we already have this with voluptuous)
2. We map the schema to code and back (serialize, deserialize).
3. A config flow would simply be _an_ interface to fill the Configuration based on that schema. The developers can do whatever they wish during their flows and fill the configuration as they want, decoupling the presentation from the data.
4. Schemas would be versioned, so a user could be presented with a flow / form / notification if and when they need to pay attention to potential configuration changes.
5. As a bonus, if no config flow is given for the integration (or even if someone prefers a form entry over the given flow), we can still display a form based on the schema, including nice UI validators based on the schema (checkbox for boolean, radio button/dropdown for enums, validating IP address fields for addresses etc).

Now, those who want to manually manage their configuration manually could simply export a copy of the generated configuration (or prepare one manually by themselves) and adjust it as they want. The manually configured configuration would then take the precedence.

Also, the examples in documentation could be generated automatically using the schema (having example values), so that would be one less thing for doc maintainers to worry about.

P.S. It may be worthwhile to take a look into how the (user-visible) configuration is done in WoT and other specifications, is there something we can adjust to our needs?

robloh:
As someone struggling to write an integration and of ease of system use & changes in the future I'm really liking @rytilahti proposal

balloob:
Let's not derail the conversation with a proposal for something that is already in place with config flows, which work with the existing voluptuous schemas. You can open a new issue if you want to discuss this.

ciotlosm:
I'm not sure if this is still being discussed, but I've been away for most of the "config flow" implementation and I thought my experience getting back into this would be helpful. 

I really like config flows for all the benefits it brings with area registry and entity registry. Changing things without restarts speeds up the setup incredibly.

 I used MQTT config flow integration renaming things as I was sitting close by on the fly (no restarts!). I was able to setup automation sin the UI from my mobile phone (!) while around the house setting up devices discovered on the fly through the mqtt integration. That was amazing to experience. Good luck doing that for bigger home setups without a laptop at hand when you need to tinker with yaml and home assistant restarts. 

There are three worlds as I see:
1. Users that want it simple and advanced users  wanting it easy for friends they set this up for to avoid maintenance
2. Advanced users that find the ability to use `vi` and `git` useful and cool
3. Developers (usually advanced users) that want to contribute as easily as possible

1. and 3. are the opposing side of the spectrum and config flow favors 1 while config yaml favors 3. 

Reading through the comments it seems that balloob already built something to make config flows easier for 3.

I do think that for higher adoption of Home Assistant going with config flow as default is the best approach. This makes it even easier to maintain remote installation for home assistant for friends without special addons to reach the config yaml. 

One thing that I find missing that would help a lot and will reduce confusion is a "simple" version of the Integrations page that just shows integration that have a config flow. Something similar to how the implementation of 'advanced' mode was added in the UI, but I guess this is separate to this discussion. 

There are probably users that would just be happy to select backup of the configuration by integration and remove the anxiety that they no longer have the yaml at hand.

MartinHjelmare:
I think the current consensus is:

1. Config yaml, if used, should be centralized under the integration domain. Platform yaml sections should be considered legacy for new integrations.
2. Config yaml should be optional. The contributor, or code owner, decides if config yaml should be implemented.
3. The promise to users is that for home assistant core config yaml will stay but for integrations it's up to each integration to support or not support config yaml.

Things that don't have full consensus or clear path forward:

1. Should we require config flow? We now have scaffolding scripts for config flow. :rocket: Import of config yaml and keeping the config entry updated when config yaml is still not clearly generally solved.
2. Should we allow removing config yaml from existing integration and if so how should that be decided? Is it up to the code owner? Should we require approval in an architecture issue or similar? Can we trust integrations to self manage this, or do we need more central control?

MartinHjelmare:
I'll close this now and open a new ADR PR for point 1 above where we have consensus.

PR #339, opened 2020-01-23
Suggest Entity.entity_registry_enabled_default

balloob:
Small update to ADR003 to suggest using `Entity.entity_registry_enabled_default`

### DIFF
--- a/adr/0003-monitor-condition-and-data-selectors.md
+++ b/adr/0003-monitor-condition-and-data-selectors.md
@@ -27,7 +27,7 @@ Layer model:
   -------------      ---------      ------------
 ```
 
-Integrations can set the `entity_registry_enabled_default` property on their entity objects to instruct Home Assistant to disable certain entities by default.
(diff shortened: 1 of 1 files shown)


adr/0003-monitor-condition-and-data-selectors.md
```diff
@@ -27,6 +27,8 @@ Layer model:
   -------------      ---------      ------------
 ```
 
+Integrations can set the `entity_registry_enabled_default` property on their entity objects to instruct Home Assistant to disable certain entities by default.
```
rytilahti: TBF, I don't understand how this works, but shouldn't this be named `disabled` if it's used to disable conditions? Can we have a more concrete example of how this is used / implemented by integration developers?

Kane610: There are multiple implementations in the code base should we reference integrations using this functionality?

balloob: The first time that an entity is added to the entity registry, a user can return `False` from this property (default is `True`) and it will add the entity to the registry, but disable it. The user can then later enable it in the entity registry if they care about this data.

PR #343, opened 2020-02-06
ADR 0009: Do not remove YAML keys

balloob:
See file.

### DIFF
--- a/adr/0009-do-not-remove-yaml-keys.md
+++ b/adr/0009-do-not-remove-yaml-keys.md
@@ -1,0 +1,21 @@
+# 0009. Do not remove YAML config keys.
+
+Date: 2020-02-06
+
+## Status
+
+Accepted
+
+## Context
+
+When integrations evolve, config keys can change. This means that options that are currently valid can become invalid in a new version.
+
+## Decision
+
+When a config key is no longer used in a configuration schema, don't remove it but instead wrap it in `cv.deprecated`. The user will get a warning and that is fine.
+
+## Consequences
+
+Reduce the number of breaking changes and instead have more deprecation warnings.
+
+Remove the version number from `cv.deprecated` validator.


elupus:
So what about the function linked to the config?

elupus:
What I meant is, if function is removed linked to the config, it's just as much a breaking change. Just more hidden in a sense.

balloob:
It's a breaking change, but your Home Assistant instance will still start if you have it in your config.

elupus:
The removed in version, is still beneficial, it doesn't need to be a hard fault though, can remain a warning. But it serves as a indication to devs when function can be removed.

pvizeli:
We should also write a timeframe

tribut:
In some cases, such as when a `password` key or similar is deprecated (renamed/moved or removed due to a move to some different form of authentication, ...) it may actually beneficial not to start HA. However, this might be an orthogonal discussion.

balloob:
We should really really try to start Home Assistant. In some cases people use Home Assistant to access their editors etc. If we can make Home Assistant start and just let them know something is not working, that's a lot nicer.

MartinHjelmare:
We're pushing a lot of responsibility on the integration developers with this decision. Can't we remove the failing integrations from config and start home assistant with just passing integrations?

That way users can upgrade and read the docs and use the config validator as they normally do.

elupus:
Isn't what @martinhjelmare say what we actually do in a Hass core startup. Isnt the upgrade mechanism just a bit overzelus in the rollback?

balloob:
That is a very good point @elupus and @MartinHjelmare 

Maybe instead we should have the Supervisor still allow restarts and upgrades, as long as the YAML can be processed.

frenck:
For Home Assistant restarts, I think, now with Safe Mode, we can think of removing it. Nevertheless, we could make restarts Core version aware, to still catch people off guard running older versions of the Core.

For upgrades, it doesn't check the config. It just upgrades. If the core fails to start/respond, it will rollback.

balloob:
Hmm chasing ghosts

pvizeli:
I think also, we should just remove failing components on start. If we add a deprication warning and the user update only twice per year, it will have the same issue if we not solve that.

PR #361, opened 2020-04-05
Add ADR 0007: Config yaml structure for integrations

### LINKED ISSUE #142: Config structure for embedded platforms

MartinHjelmare:
Closes #142 

This is broken out from #289. See that PR for the discussion.

### DIFF
--- a/adr/0007-integration-config-yaml-structure.md
+++ b/adr/0007-integration-config-yaml-structure.md
@@ -12,7 +12,7 @@ Accepted
 
 There are currently many different ways of structuring an integration and its config in Home Assistant. We allow config YAML, either under the integration key or under platform keys with the integration name. We allow a config flow with configuration entered via the GUI, stored in config entries. We allow importing config YAML to a config entry via config flow.
 
-This ADR focusses on the configuration YAML structure and its use in integrations.
+This ADR focuses on the configuration YAML structure and its use in integrations.
 
 The many options for configuration impact how the integration is structured and what Home Assistant backend APIs are used.
 
--- a/adr/0007-integration-config-yaml-structure.md
+++ b/adr/0007-integration-config-yaml-structure.md
@@ -1,0 +1,42 @@
+# 7. Config yaml structure for integrations
+
+Date: 2020-04-05
+
+## Status
+
+Accepted
+
+---
+
+## Context
+
+There are currently many different ways of structuring an integration and its config in Home Assistant. We allow config yaml, either under the integration key, or under platform keys with the integration name. We allow a config flow with configuration entered via the GUI, stored in config entries. We allow importing config yaml to a config entry via config flow.
+
+This ADR will only focus on the configuration yaml structure and use in integrations.
+
+The many options for configuration impacts how the integration is structured and what Home Assistant backend APIs are used.
+
+1. Integrations that only have a single platform puts the configuration under a platform key and just uses `async_setup_platform` to set up the platform.
+2. Integrations that have many platforms sometimes centralize the configuration to the integration domain name key and loads platforms via `discovery.async_load_platform`, which in turn calls `async_setup_platform`.
+3. If the integration has multiple platforms but doesn't have a centralized config, the user needs to enter config under the integration domain to set up the integration and each platform key to set up each platform. This loads each platform via `async_setup_platform`.
+
+All backend APIs in point 1-3 above also have sync versions, to further increase the options.
+
+This multitude of options for config and integration structure makes it:
+
+A. Harder for users to know how to configure the integrations.
+B. Harder for contributors to know what way to implement and support.
+C. Harder for Home Assistant to support confused users and contributors and harder to support and extend a disparate code base.
+
+---
+
+## Decision
+
+We limit the configuration yaml structure to one way for new integrations. We require all yaml config, if present, for an integration to be located under the integration domain key in config yaml, for all new integrations (point 2 above).
+
+## Consequences
+
+- New integrations have to put the integration configuration section under the integration domain key if configuration yaml is used.
+- Integration platforms have to be loaded via `discovery.async_load_platform` or `discovery.load_platform` if config entries is not used. This is a small increase in complexity for new integration contributions.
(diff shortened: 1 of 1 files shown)


adr/0007-integration-config-yaml-structure.md
```diff
@@ -0,0 +1,53 @@
...
+2. Integrations that have many platforms sometimes centralize the configuration to the integration domain name key and loads platforms via `discovery.async_load_platform`, which in turn calls `async_setup_platform`.
+3. If the integration has multiple platforms but doesn't have a centralized config, the user needs to enter config under the integration domain to set up the integration and each platform key to set up each platform. This loads each platform via `async_setup_platform`.
+
+All backend APIs in point 1-3 above also have sync versions, to further increase the options.
+
+This multitude of options for config and integration structure makes it:
+
+A. Harder for users to know how to configure the integrations.
+B. Harder for contributors to know what way to implement and support.
+C. Harder for Home Assistant to support confused users and contributors and harder to support and extend a disparate code base.
```
frenck, line 29:
```suggestion
There are currently many different ways of structuring an integration and its config in Home Assistant. We allow config YAML, either under the integration key or under platform keys with the integration name. We allow a config flow with configuration entered via the GUI, stored in config entries. We allow importing config YAML to a config entry via config flow.

This ADR focusses on the configuration YAML structure and its use in integrations.

The many options for configuration impact how the integration is structured and what Home Assistant backend APIs are used.

1. Integrations that only have a single platform, place the configuration under a platform key and use `async_setup_platform` to set up the platform.
2. Integrations that have multiple platforms, sometimes centralize the configuration to the integration domain name key and load platforms via `discovery.async_load_platform`, which in turn calls `async_setup_platform`.
3. If the integration has multiple platforms but doesn't have a centralized config, the user needs to add configuration under the integration domain to set up the integration and each platform key to set up each platform. This loads each platform via `async_setup_platform`.

All backend APIs in point 1-3 above also have sync versions, to further increase the options.

This multitude of options for config and integration structure makes it:

A. Harder for users to know how to configure the integrations.
B. Harder for contributors to know what way to implement and support.
C. Harder for Home Assistant to support confused users and contributors and harder to support and extend a disparate code base.
```

adr/0007-integration-config-yaml-structure.md
```diff
@@ -0,0 +1,53 @@
+# 7. Config yaml structure for integrations
```
frenck, line 1:
```suggestion
# 7. Config YAML structure for integrations
```

adr/0007-integration-config-yaml-structure.md
```diff
@@ -0,0 +1,53 @@
...
+```yaml
+# Allowed
+awesome_integration:
+  username: user
+
+# Not allowed
+sensor:
+  - platform: awesome_integration
+    username: user
+```
```
frenck, line 46:
```suggestion
## Decision

We limit the configuration YAML structure to one way for new integrations. We require all YAML configuration, if present, for an integration to be located under the integration domain key in configuration YAML, for all new integrations (point 2 above).

```yaml
# Allowed
awesome_integration:
  username: user

# Not allowed
sensor:
  - platform: awesome_integration
    username: user
```
```

File: adr/0007-integration-config-yaml-structure.md [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+
+## Consequences
+
+- New integrations have to put the integration configuration section under the integration domain key if configuration yaml is used.
+- Integration platforms have to be loaded via `discovery.async_load_platform` or `discovery.load_platform` if config entries is not used. This is a small increase in complexity for new integration contributions.
+- We will decrease breaking changes when an integration wants to add more platforms and move the configuration from a per platform configuration to a centralized integration domain configuration section.
+- We will increase the consolidation of Home Assistant Core configuration yaml integration structure instead of increasing disparity.
```
frenck, line 53:
```suggestion
## Consequences

- New integrations have to put the integration configuration section under the integration domain key if configuration YAML is used.
- Integration platforms have to be loaded via `discovery.async_load_platform` or `discovery.load_platform` if not using configuration entries. This is a small increase in complexity for new integration contributions.
- Decreases the number of breaking changes when an integration wants to add more platforms and move the configuration from a per platform configuration to a centralized integration domain configuration section.
- Increases the consolidation of Home Assistant Core configuration YAML integration structure; instead of increasing disparity.
```

frenck:
Just thought about this a little more. While we are not accepting any new ones after this has been merged. Any thoughts about the existing single platform integrations? E.g., do we accept changes on those? Or do we require them to be migrated when new features are introduced? (excluding bug fixes from that).

MartinHjelmare:
> 1. Do we require existing single platform integrations to be migrated when new features are introduced?
> 2. What about currently open PRs that are violating this?

1. I think we should not accept changes for the config, but other changes could be ok.
2. If we haven't started review of the PR, we'll require them to update. If we have started review I think it's ok to merge as is.

adr/0007-integration-config-yaml-structure.md
```diff
@@ -0,0 +1,52 @@
...
+
+Accepted
+
+---
+
+## Context
+
+There are currently many different ways of structuring an integration and its config in Home Assistant. We allow config YAML, either under the integration key or under platform keys with the integration name. We allow a config flow with configuration entered via the GUI, stored in config entries. We allow importing config YAML to a config entry via config flow.
+
+This ADR focusses on the configuration YAML structure and its use in integrations.
```
MartinHjelmare, line 15:
```suggestion
This ADR focuses on the configuration YAML structure and its use in integrations.
```

PR #363, opened 2020-04-09
Translations 2.0

balloob:
See ADR File.

### DIFF
--- a/adr/0009-translations-2.0.md
+++ b/adr/0009-translations-2.0.md
@@ -71,6 +71,7 @@ We should drop the `title` from `strings.json` if the integration is for a produ
   }
   ```
 - Import all existing state attribute keys + translations from frontend
+- Replace as many existing translations as possible with references
 
 #### Frontend
 Load part of backend translations when necessary.
--- a/adr/0009-translations-2.0.md
+++ b/adr/0009-translations-2.0.md
@@ -72,6 +72,7 @@ We should drop the `title` from `strings.json` if the integration is for a produ
   ```
 - Import all existing state attribute keys + translations from frontend
 - Replace as many existing translations as possible with references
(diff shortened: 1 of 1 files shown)


adr/0009-translations-2.0.md
```diff
@@ -0,0 +1,94 @@
...
+          },
+        }
+      }
+    }
+    ```
+- Import all existing state keys + translations from frontend
+- Add a new `state_attributes` to `strings.json` to allow offering translations for state attributes.
+  ```json
+  {
+    "state_attributes": {
```
balloob, line 59: This missses the translation label for `hvac_action`.

adr/0009-translations-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+- Remove built-in translations for state attributes (`state_attributes`) and use backend provided strings instead.
+- Adopt cleaning keys script from backend ([see this PR](https://github.com/home-assistant/core/pull/33802))
+- Remove all state and state attributes keys from translations once imported into backend
+
+## Consequences
+
+- Frontend will load faster and use less memory
+- Couple translations of integrations to the integration instead of the frontend release
+
+The `moon` and `season` integrations will no longer have their entities translated. I suggest that if we see a need to bring this back, we could consider migrating `moon` and `season` to not write sensor entities but instead have them write states under their own domain (`moon.moon`) like the `sun` integration does. They can then add `state` to their own `strings.json` which can then be loaded when the frontend loads.
```
pvizeli, line 95: As user of moon and season, I agree, that are no sensors and we should migrate this as part of this ADR change and not suggest it. So we have `sun`, `moon`, `season` which also should provide later device automation and condition. That stuff is really useful for automation (okay, moon is less as season).

PR #374, opened 2020-04-27
ADR-0011 - Discovery Requires Unique ID

ctalkington:
To protect project goals and to provide clarity to our users and contributors,
we're introducing the following rules on how discoverable integrations need to be designed:

- Integrations that are discoverable must provide an unique id (via `async_set_unique_id`) 
to allow the user to ignore the discovered config entry.

These rules apply to all new integrations. Existing integrations should be reviewed 
for possible ways to improve identification during discovery.

### DIFF
--- a/adr/0011-discovery-requires-unique-id.md
+++ b/adr/0011-discovery-requires-unique-id.md
@@ -8,27 +8,39 @@ Accepted
 
 ## Context
 
-There is a mechanism to ignore discovered config flows using unique ids attached to config flow but not all integrations may set a unique id due to lack of truly unique identifiers or just not having been updated for concerns of backwards compatibility.
+There is a mechanism to ignore discovered config flows using unique ids attached to config flow
+but not all integrations may set a unique id due to lack of truly unique identifiers
+or just not having been updated for concerns of backwards compatibility.
 
-This creates an UX issue that has started to become more pronounced with more config flows being added every release (which is awesome!). The real issue is not all discovery processes are created equal and that becomes an issue for consistency.
+This creates an UX issue that has started to become more pronounced with more config flows
+being added every release (which is awesome!). The real issue is not all discovery processes 
+are created equal and that becomes an issue for consistency.
 
 ## Proposal
 
-All future integratons with discoverable config flows should be required to set an unique id during discovery.
+All future integratons with discoverable config flows should be required to set an
+unique id during discovery.
 
-All existing integrations should be reviewed for possible ways to improve identification during discovery config flow. A whitelist could be developed to allow existing integrations to remain functional under new requirements.
+All existing integrations should be reviewed for possible ways to improve identification
+during discovery config flow. A whitelist could be developed to allow existing integrations
+to remain functional under new requirements.
 
-If a unique id can sometimes not be available such as older devices lack identifiers in zeroconf/ssdp/etc then the device should be prevented from being shown to user.
+If a unique id can sometimes not be available such as older devices lack identifiers in 
+zeroconf/ssdp/etc then the device should be prevented from being shown to user.
 
 ## Decision
 
 To protect project goals and to provide clarity to our users and contributors,
-we’re introducing the following rules on how discoverable integrations need to be configured:
+we're introducing the following rules on how discoverable integrations need to be configured:
 
-- Integrations that are discoverable must provide an unique id (via `async_set_unique_id`) to allow the user to ignore the discovered config entry.
+- Integrations that are discoverable must provide an unique id (via `async_set_unique_id`) 
+to allow the user to ignore the discovered config entry.
 
-These rules apply to all new integrations. Existing integrations should be reviewed for possible ways to improve identification during discovery.
+These rules apply to all new integrations. Existing integrations should be reviewed 
+for possible ways to improve identification during discovery.
 
 ## Consequences
 
-Discovery UX improves at the expense of some unidentifiable or non-conforming devices requiring manual setup.
+- Removes confusion and questions around the ability to ignore unwanted discoveries
+  for all users.
+- This might impact the number of integrations that are able to be discovered.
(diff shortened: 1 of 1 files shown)


adr/0011-discovery-requires-unique-id.md
```diff
@@ -0,0 +1,45 @@
...
+
+Accepted
+
+## Context
+
+There is a mechanism to ignore discovered config flows using unique ids attached to config flow
+but not all integrations may set an unique id due to lack of truly unique identifiers
+or just not having been updated for concerns of backwards compatibility.
+
+This creates an UX issue that has started to become more pronounced with more config flows
```
MartinHjelmare, line 15:
```suggestion
This creates a UX issue that has started to become more pronounced with more config flows
```

adr/0011-discovery-requires-unique-id.md
```diff
@@ -0,0 +1,45 @@
...
+being added every release (which is awesome!). The real issue is not all discovery processes 
+are created equal and that becomes an issue for consistency.
+
+## Proposal
+
+All new integratons with discoverable config flows should be required to set an
```
MartinHjelmare, line 21:
```suggestion
All new integratons with discoverable config flows should be required to set a
```

adr/0011-discovery-requires-unique-id.md
```diff
@@ -0,0 +1,45 @@
...
+unique id during discovery.
+
+All existing integrations with discoverable config flows should be reviewed for
+possible ways to improve identification during discovery. A whitelist could be developed
+to allow existing integrations to remain functional under new requirements.
+
+If an unique id can sometimes not be available such as older devices that lack unique identifiers
+in zeroconf/ssdp/api/etc then the config entry should not be shown to user.
```
MartinHjelmare, line 29:
```suggestion
If a unique id can not be made available, such as on older devices that lack unique identifiers
in zeroconf/ssdp/api/etc, then the config entry should not be shown to user.
```

adr/0011-discovery-requires-unique-id.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Decision
+
+To protect project goals and to provide clarity to our users and contributors,
+we're introducing the following rules on how discoverable integrations need to be designed:
```
MartinHjelmare, line 34: I think we can drop this. It seems to be copied from ADR 10.

adr/0011-discovery-requires-unique-id.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+There is a mechanism to ignore discovered config flows using unique ids attached to config flow
+but not all integrations may set an unique id due to lack of truly unique identifiers
+or just not having been updated for concerns of backwards compatibility.
```
frenck, line 13:
```suggestion
or just not having been updated for concerns of backward compatibility.
```

adr/0011-discovery-requires-unique-id.md
```diff
@@ -0,0 +1,45 @@
...
+
+This creates an UX issue that has started to become more pronounced with more config flows
+being added every release (which is awesome!). The real issue is not all discovery processes
```
frenck, line 16:
```suggestion
being added every release (which is awesome!). The real issue is that not all discovery processes 
```

adr/0011-discovery-requires-unique-id.md
frenck: I've added your suggestion, as "integrations" was misspelled here as well.

adr/0011-discovery-requires-unique-id.md
frenck: Edited your suggestion, as SSDP, API should be capitalized as well and `the` was missing before `user`.

ctalkington:
i think i got all the suggestions accepted here

adr/0011-discovery-requires-unique-id.md
```diff
@@ -0,0 +1,43 @@
...
+All existing integrations with discoverable config flows should be reviewed for
+possible ways to improve identification during discovery. A whitelist could be developed
+to allow existing integrations to remain functional under new requirements.
+
+If a unique id can not be made available, such as on older devices that lack unique identifiers
+in zeroconf/SSDP/API/etc, then the config entry should not be shown to the user.
+
+## Decision
+
+Integrations that are discoverable must provide an unique id (via `async_set_unique_id`)
```
MartinHjelmare, line 34:
```suggestion
Integrations that are discoverable must provide a unique id (via `async_set_unique_id`) 
```

scop:
Not sure what's the best place to discuss existing integrations, but I'll just do a quick huawei_lte braindump here:

We are currently not using anything from discovery as as unique id because upnp is optional for these devices, and might be later turned off by the user which I believe would result in device identity disassociation in HA.

Due to a similar reason we are not using the device serial number as unique id either -- the serial is available only when using a username/password to access the device, not when using unauthenticated mode, and the user may want to switch to it later.

So we currently use the device mac, which has drawbacks but at least works consistently. Hum well except it has similar issues, it'll change if the device moves to/from behind a gateway. Sigh. Anyway that's where we are at the moment.

balloob:
If upnp is turned off, the device will also not be discovered 👍

scop:
Yep, that's why we can't (and don't) use anything from the discovered data as the unique id.

balloob:
It's the opposite actually :) See, config entry unique ID is only used to ignore discovered config entries. So using it is fine.

scop:
But we need a unique id for other things unrelated to discovery, and we cannot assume that the device will be discovered in the first place (upnp might never be turned on).

elupus:
The unique ID of the config entry doesn't need to be same as entities.

scop:
Ah! Missed that completely.

PR #392, opened 2020-06-03
ADR: Define supported installation method

balloob:
ADR to define what is a supported installation method.

As introduced earlier in https://www.home-assistant.io/blog/2020/05/26/installation-methods-and-community-guides-wiki/

### DIFF
--- a/adr/0012-define-supported-installation-method.md
+++ b/adr/0012-define-supported-installation-method.md
@@ -1,0 +1,33 @@
+# 0012. Define supported installation method
+
+Date: 2020-06-03
+
+## Status
+
+Accepted
+
+## Context
+
+If you look at our documentation, it’s all over the place. Install it in Docker, in a VM, on a NAS or on one of the many Linux distributions.
+
+The reason we have this many guides is that since the start of the Home Assistant website, we have always gladly accepted every contribution to get Home Assistant running on any platform. The more the merrier!
+
+However, in software, nothing ever stays the same. All software gets updates to fix bugs, fix security vulnerabilities, improve performance or to add new features. As a software application you need to grow along or else you get stuck with an insecure system.
+
+So as Home Assistant grows and evolves, some of these installation guides became outdated and stopped working. We wouldn’t even know it was broken until a user raised an issue. But when they do, we wouldn’t know how to fix it unless we could get a hold of the original contributor.
+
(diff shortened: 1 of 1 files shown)


adr/0012-define-supported-installation-method.md
```diff
@@ -0,0 +1,33 @@
...
+
+So as Home Assistant grows and evolves, some of these installation guides became outdated and stopped working. We wouldn’t even know it was broken until a user raised an issue. But when they do, we wouldn’t know how to fix it unless we could get a hold of the original contributor.
+
+This can be frustrating. Any guide on our official website should lead to a working system. A system that not only works today, but also tomorrow.
+
+## Decision
+
+A supported installation method in the Home Assistant context means:
+
+> A way of installing and running Home Assistant One, Home Assistant Container, Home Assistant Supervised or Home Assistant Core; in a way that is supported by the Home Assistant developers. Supported means the installation method is tested and documented in the official documentation. Running Home Assistant using such a supported method, leads to the optimal user experience.
```
Adminiuga, line 25: > A way of installing and running Home Assistant One, Home Assistant Container, Home Assistant Supervised or Home Assistant Core

What is "Home Assistant One" ?

balloob: 😆 

That is for the next ADR, but that will be the new name of referring to the installation method of what is currently Home Assistant. 

The idea behind it is that it is confusing when people have issues. What are you running? "Home Assistant". "Ok, but what HA?"

Adminiuga: One to  rule them all :)

DubhAd, line 25: > Home Assistant One
is this the latest name for the thing previously called Hass.io?

thomasloven, line 25: I think this introduces another question.
What is *installed* and what is the *method*?

As I see it, at the base there's Home Assistant - the python program.
Then there's different ways of making that run
- One
- in a container
- in a container that's controlled by the supervisor
- standalone

That would make One, Container, Supervised and - in a way - Core *installation methods*, which makes this a circular definition.

balloob: Every method installs core, which is what is the home automation part. But how it is installed and managed is part of the installation method, which is what we're defining/naming.

PR #394, opened 2020-06-03
ADR: Home Assistant Supervised as supported installation method

balloob:
Add an ADR for adding Home Assistant Supervised as a supported installation method as defined in #392.

### DIFF
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -12,11 +12,9 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 ## Decision
 
-This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor.
+This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. This system will run the Home Assistant Supervisor. The Supervisor is not just an application, it is a full appliance that manages the whole system. It will clean up, repair or reset settings to default if they no longer match expected values.
 
-The Supervisor is not just an application, it is a full appliance that manages the whole system. It will clean up, repair or reset settings to default if they no longer match expected values.
-
-By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or correcting a wrong configuration.
+By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. Home Assistant is not responsible for issues that come up with your host or config/resource conflicts caused by other applications installed on the host system.
 
 This method is considered advanced and should only be used if one is an expert in managing a Linux operating system, Docker and networking.
 
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -28,8 +28,8 @@ Docker CE (Community Edition) is the only supported containerization method for 
 - Systemd >= 239
 - NetworkManager >= 1.18.0
 - Avahi >= 0.7
-- AppArmor == 2.13.x
-- Debian Linux Debian 10 aka Buster
+- AppArmor == 2.13.x (built into the kernel)
+- Debian Linux Debian 10 aka Buster (no derivatives)
 
 Only the above-listed version of Debian Linux is supported for running this installation method. When a new major version of Debian is released, the previous major version is dropped, with a deprecation time of 4 months. An exception to this rule occurs if the new version does not meet the requirements of the Supervisor.
 
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -1,4 +1,4 @@
-# 0012. Installation method: Home Assistant Container
+# 0014. Installation method: Home Assistant Supervised
 
 Date: 2020-06-03
 
@@ -14,13 +14,13 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor. You need to understand that the Supervisor is not just an application, it is a full appliance.
 
-By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or wrong configuration.
+By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or correcting a wrong configuration.
 
 This method is considered advanced and should only be used if one is an expert in managing a Linux operating system, Docker and networking.
 
 ### Supported Operating System, System dependencies and versions
 
-Docker CE (Community Edition) is the only supported containerization method for Home Assistant Supervised. We only support FSH 3.0 on the host file system.
+Docker CE (Community Edition) is the only supported containerization method for Home Assistant Supervised. We only support FHS 3.0 on the host file system.
 
 - Docker CE >= 19.03
 - Systemd >= 239
@@ -29,7 +29,7 @@ Docker CE (Community Edition) is the only supported containerization method for 
 - AppArmor == 2.13.x
 - Debian Linux Debian 10 aka Buster
 
-Only the above-listed version of Debian Linux is supported for running this installation method. When a new major version of Debian is released, the previous major version is dropped, with a deprecation time of 4 months. An exception to this rule occurs if the new version does not solve the requirements of the Supervisor.
+Only the above-listed version of Debian Linux is supported for running this installation method. When a new major version of Debian is released, the previous major version is dropped, with a deprecation time of 4 months. An exception to this rule occurs if the new version does not meet the requirements of the Supervisor.
 
 ### Additional supported conditions
 
@@ -53,14 +53,14 @@ In case any abnormality is detected that prevents Home Assistant from functionin
 * **Access USB devices:** done by the installer
 
 * **Maintenance Home Assistant installation**
-  Home Assistant can be maintained from the supervisor. Includes rollback when the update fails.
+  Home Assistant can be maintained from the Supervisor. Includes rollback when the update fails.
 
   - **Python upgrades:** Included in Home Assistant updates
   - **Installing Python dependencies:** Included in Home Assistant updates
   - **Updating Home Assistant:** Via the UI
 
 - **Maintenance Operating System**
-  The user is responsible for maintaining the operating system. Since it is a supervised installation, the user is also responsible for updating the components that are required by the supervisor. The user is also responsible for installing or changing anything on their system that will interfere with the Supervised installation. Examples are software that will update Docker containers managed by the Supervisor.
+  The user is responsible for maintaining the operating system. Since it is a supervised installation, the user is also responsible for updating the components that are required by the Supervisor. The user is also responsible for not installing or changing anything on their system that will interfere with the Supervised installation. Examples are software that will update Docker containers managed by the Supervisor.
 
   - **Security updates for OS:** Responsibility of the user.
   - **Maintaining the components required for the Supervisor:** Responsibility of the user. Over time as Supervisor requirements change, you might have to upgrade your OS to be able to use the required version.
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -48,26 +48,26 @@ In case any abnormality is detected that prevents Home Assistant from functionin
 ### Required Expertise
 
 - **Installation**
-  The user first needs to install a Linux-based operating system (i.e. Debian). They then need to run the installer script.
+  The user first needs to install Debian 10 and make sure all the required components are installed and are the correct version. They then need to run the installer script.
 
 * **Start when the system is started:** done by the installer
 * **Run with full network access:** done by the installer
 * **Access USB devices:** done by the installer
 
-* **Maintenance Home Assistant installation**
-  Home Assistant can be maintained from the Supervisor. Includes rollback when the update fails.
+* **Maintaining the Home Assistant installation**
+  Home Assistant can be maintained from the Supervisor. This includes a rollback when the update fails.
 
   - **Python upgrades:** Included in Home Assistant updates
   - **Installing Python dependencies:** Included in Home Assistant updates
   - **Updating Home Assistant:** Via the UI
 
-- **Maintenance Operating System**
+- **Maintaining the Operating System**
   The user is responsible for maintaining the operating system. Since it is a supervised installation, the user is also responsible for updating the components that are required by the Supervisor. The user is also responsible for not installing or changing anything on their system that will interfere with the Supervised installation. Examples are software that will update Docker containers managed by the Supervisor.
 
   - **Security updates for OS:** Responsibility of the user.
   - **Maintaining the components required for the Supervisor:** Responsibility of the user. Over time as Supervisor requirements change, you might have to upgrade your OS to be able to use the required version.
 
-**Conclusion:** Expert. Maintaining a Linux operating system to a very specific set of requirements is hard.
+**Conclusion:** Expert. Maintaining a Debian 10 installation to a very specific set of requirements is hard.
 
 ## Consequences
 
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -1,0 +1,78 @@
+# 0012. Installation method: Home Assistant Container
+
+Date: 2020-06-03
+
+## Status
+
+Accepted
+
+## Context
+
+Define a supported installation method as per [ADR-0012](https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md).
+
+## Decision
+
+This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor. You need to understand that the Supervisor is not just an application, it is a full appliance.
+
+By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or wrong configuration.
+
+This method is considered advanced and should only be used if one is an expert in managing a Linux operating system, Docker and networking.
+
+### Supported Operating System, System dependencies and versions
+
+Docker CE (Community Edition) is the only supported containerization method for Home Assistant Supervised. We only support FSH 3.0 on the host file system.
+
+- Docker CE >= 19.03
+- Systemd >= 239
+- NetworkManager >= 1.18.0
+- Avahi >= 0.7
+- AppArmor == 2.13.x
+- Debian Linux Debian 10 aka Buster
+
+Only the above-listed version of Debian Linux is supported for running this installation method. When a new major version of Debian is released, the previous major version is dropped, with a deprecation time of 4 months. An exception to this rule occurs if the new version does not solve the requirements of the Supervisor.
+
+### Additional supported conditions
+
+This installation method can easily be broken if one manages the operating system incorrectly. Therefore, the following additional conditions apply:
+
+- The operating system is dedicated to running Home Assistant Supervised.
+- All system dependencies are installed according to the manual.
+- No additional software, outside of the Home Assistant ecosystem, is installed.
+- Docker needs to be correctly configured to use overlayfs2 storage and journald as the logging driver.
+- NetworkManager is installed but not actively used.
+
+In case any abnormality is detected that prevents Home Assistant from functioning, the Home Assistant Supervisor will report this to the user and block updates to prevent installations from breaking.
+
+### Required Expertise
+
+- **Installation**
+  The user first needs to install a Linux-based operating system (i.e. Debian). They then need to run the installer script.
+
+* **Start when the system is started:** done by the installer
+* **Run with full network access:** done by the installer
+* **Access USB devices:** done by the installer
+
+* **Maintenance Home Assistant installation**
+  Home Assistant can be maintained from the supervisor. Includes rollback when the update fails.
+
+  - **Python upgrades:** Included in Home Assistant updates
+  - **Installing Python dependencies:** Included in Home Assistant updates
+  - **Updating Home Assistant:** Via the UI
+
+- **Maintenance Operating System**
+  The user is responsible for maintaining the operating system. Since it is a supervised installation, the user is also responsible for updating the components that are required by the supervisor. The user is also responsible for installing or changing anything on their system that will interfere with the Supervised installation. Examples are software that will update Docker containers managed by the Supervisor.
+
+  - **Security updates for OS:** Responsibility of the user.
+  - **Maintaining the components required for the Supervisor:** Responsibility of the user. Over time as Supervisor requirements change, you might have to upgrade your OS to be able to use the required version.
+
+**Conclusion:** Expert. Maintaining a Linux operating system to a very specific set of requirements is hard.
+
+## Consequences
+
+Update documentation on how to install this method, the required experience and the expected maintenance.
+
+Move existing related documentation that does not match the requirements to the community guides wiki.
+
+Notify user during onboarding of expected maintenance for their installation method.
+
+Add checks to Home Assistant to warn (but not forbid) when not all requirements are met for this installation method.
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -14,6 +14,8 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor.
 
+The Supervisor is not just an application, it is a full appliance that manages the whole system. It will clean up, repair or reset settings to default if they no longer match expected values.
+
 By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or correcting a wrong configuration.
 
 This method is considered advanced and should only be used if one is an expert in managing a Linux operating system, Docker and networking.
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -12,7 +12,7 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 ## Decision
 
-This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor. You need to understand that the Supervisor is not just an application, it is a full appliance.
+This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor.
 
 By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or correcting a wrong configuration.
 
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -14,7 +14,7 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. This system will run the Home Assistant Supervisor. The Supervisor is not just an application, it is a full appliance that manages the whole system. It will clean up, repair or reset settings to default if they no longer match expected values.
 
-By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. Home Assistant is not responsible for issues that come up with your host or config/resource conflicts caused by other applications installed on the host system.
+By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. Home Assistant Supervised is provided as-is as a foundation for community supported do-it-yourself solutions. We only accept bug reports for issues that have been reproduced on a freshly installed, fully updated Debian with no additional packages.
 
 This method is considered advanced and should only be used if one is an expert in managing a Linux operating system, Docker and networking.
 


adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,78 @@
...
+
+Only the above-listed version of Debian Linux is supported for running this installation method. When a new major version of Debian is released, the previous major version is dropped, with a deprecation time of 4 months. An exception to this rule occurs if the new version does not solve the requirements of the Supervisor.
+
+### Additional supported conditions
+
+This installation method can easily be broken if one manages the operating system incorrectly. Therefore, the following additional conditions apply:
+
+- The operating system is dedicated to running Home Assistant Supervised.
+- All system dependencies are installed according to the manual.
+- No additional software, outside of the Home Assistant ecosystem, is installed.
```
balloob, line 40: This one will need some extra explaining. It will be along the lines: 

The reason for this requirement is that the Supervisor turns your system into an appliance: a system with the specific task of running Home Assistant. This is not an "app" that you install. Your system becomes Home Assistant Supervised.

This means that the supervisor will expect to be able to clean and correct the system to ensure smooth operations. Examples:

 - If a config can't be parsed, we should be able to reset it to defaults.
 - cleanup tasks will remove all Docker containers that are not running

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,78 @@
...
+
+This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor. You need to understand that the Supervisor is not just an application, it is a full appliance.
+
+By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or wrong configuration.
+
+This method is considered advanced and should only be used if one is an expert in managing a Linux operating system, Docker and networking.
+
+### Supported Operating System, System dependencies and versions
+
+Docker CE (Community Edition) is the only supported containerization method for Home Assistant Supervised. We only support FSH 3.0 on the host file system.
```
ZephireNZ, line 23: Is this meant to be [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html)?

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Docker CE >= 19.03
+- Systemd >= 239
+- NetworkManager >= 1.18.0
+- Avahi >= 0.7
+- AppArmor == 2.13.x
+- Debian Linux Debian 10 aka Buster
+
+Only the above-listed version of Debian Linux is supported for running this installation method. When a new major version of Debian is released, the previous major version is dropped, with a deprecation time of 4 months. An exception to this rule occurs if the new version does not solve the requirements of the Supervisor.
```
DubhAd, line 32: > does not solve the requirements of 
I'd suggest something more like:
> does not meet the requirements of
or
> does not provide the requirements for

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Context
+
+Define a supported installation method as per [ADR-0012](https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md).
+
+## Decision
+
+This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor. You need to understand that the Supervisor is not just an application, it is a full appliance.
+
+By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or wrong configuration.
```
DubhAd, line 17: > or wrong configuration
I would suggest
> or corroecting a wrong configuration

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,78 @@
...
+
+* **Maintenance Home Assistant installation**
+  Home Assistant can be maintained from the supervisor. Includes rollback when the update fails.
+
+  - **Python upgrades:** Included in Home Assistant updates
+  - **Installing Python dependencies:** Included in Home Assistant updates
+  - **Updating Home Assistant:** Via the UI
+
+- **Maintenance Operating System**
+  The user is responsible for maintaining the operating system. Since it is a supervised installation, the user is also responsible for updating the components that are required by the supervisor. The user is also responsible for installing or changing anything on their system that will interfere with the Supervised installation. Examples are software that will update Docker containers managed by the Supervisor.
```
DubhAd, line 63: > The user is also responsible for installing or changing anything on their system that will interfere 
Makes it sound like they should do that ;)
> The user is also responsible for not installing or changing anything on their system that will interfere

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,78 @@
...
+
+Define a supported installation method as per [ADR-0012](https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md).
+
+## Decision
+
+This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor. You need to understand that the Supervisor is not just an application, it is a full appliance.
+
+By not using the Home Assistant Operating System, the user is responsible for making sure that all required components are installed and maintained. Required components and their versions will change over time. We are not responsible for issues that come up with your host or wrong configuration.
+
+This method is considered advanced and should only be used if one is an expert in managing a Linux operating system, Docker and networking.
```
fabaff, line 19: Don't be too specific here. The details are out-lined after.

```suggestion
This method is considered advanced and should only be used if one is an expert in managing an operating system, containerization solution and networking.
```

adr/0014-home-assistant-supervised.md
frenck: Yes, that is a typo.

adr/0014-home-assistant-supervised.md
frenck: We can generalize this that way for a supervised installation, as it can't be run on another operating system or containerization system.

andriej:
Wouldn't it be good to add, that machine with Supervised installation should have static IP and static external DNS server defined?
It would solve tons of potential problems regarding core-dns, dns-loops with adguard etc.

pvizeli:
The latest dns plugin should handle that now without big issue. It need a lot misunderstanding lag of knowledge to run into such issues now. In general it's a good idea have an static IP if you use it as DNS server in your network but not all run DNS add-ons or DHCP add-ons on it.

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,78 @@
...
+
+Accepted
+
+## Context
+
+Define a supported installation method as per [ADR-0012](https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md).
+
+## Decision
+
+This installation method provides the full Home Assistant experience on a regular operating system. This means, all components from the Home Assistant method are used, except for the Home Assistant Operating System. We don’t support using the host system for other applications than the Supervisor. You need to understand that the Supervisor is not just an application, it is a full appliance.
```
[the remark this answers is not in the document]
balloob: I've reworded it to make it more part of the definition.

amelchio:
Honestly, with all these limitations I don't really see the use case. However, I am not using this installation method. Should we get feedback from someone actually using Home Assistant Supervised before moving on?

JumpMaster:
> Honestly, with all these limitations I don't really see the use case. However, I am not using this installation method. Should we get feedback from someone actually using Home Assistant Supervised before moving on?

I used HA supervised for the last two years.  I started with Hassio in an ESX VM.  I used the Supervised method to learn about Docker and I've recently used Home Assistant to learn about KVM where I am currently running Home Assistant OS/One/Full.

It's important to know the difference between Supported and Compatible.  Home Assistant Supervised will be supported on the current Debian release using Docker CE with no other applications installed.  That does not mean it will not work on Ubuntu which is Debian based.  Ubuntu in it's state today could be considered as compatible.  But if Ubuntu 21 removes systemd or AppArmor it will become incompatible.  Home Assistant will make no changes to support Ubuntu.  If the next release of Debian removes systemd or AppArmor Home Assistant will be changed to support it.

francisp2:
From : https://github.com/home-assistant/architecture/pull/394/files/9e253a3f0c68491be4225936652bed3f4aa81953#diff-0f194ffb24a65d86e5fd43aff7e9bdc9
`
- The operating system is dedicated to running Home Assistant Supervised.
- All system dependencies are installed according to the manual.
- No additional software, outside of the Home Assistant ecosystem, is installed.
 `
Does that mean you can't run any standard debian utilities as ssh, samba, mqtt ?

pvizeli:
Ubuntu using AppArmor 3.0 was never released and is in beta for around 2 years or more. There are issues no person knows what kind of not released or documented AppArmor rules is needed to fix that.

pvizeli:
> Does that mean you can't run any standard debian utilities as ssh, samba, mqtt ?

Yes, because the system was designed to use that as an Add-on. If you don't do that as we designed it, the result and issues are up to you.

JumpMaster:
Why Docker CE over Docker IO?  On Ubuntu I've stuck to Docker.io as to get a fully supported product from Ubuntu.

My thoughts are based on the reply "Be wary of docker-ce" on this topic.
https://stackoverflow.com/questions/45023363/what-is-docker-io-in-relation-to-docker-ce-and-docker-ee

frenck:
@JumpMaster well, with the above, Ubuntu won’t be supported... 🤷‍♂️

Docker.io has caused many surprises in the past. As this is about what is supported, we of course will follow Docker, not Ubuntu’s maintained version of it.

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,80 @@
...
+- No additional software, outside of the Home Assistant ecosystem, is installed.
+- Docker needs to be correctly configured to use overlayfs2 storage and journald as the logging driver.
+- NetworkManager is installed but not actively used.
+
+In case any abnormality is detected that prevents Home Assistant from functioning, the Home Assistant Supervisor will report this to the user and block updates to prevent installations from breaking.
+
+### Required Expertise
+
+- **Installation**
+  The user first needs to install a Linux-based operating system (i.e. Debian). They then need to run the installer script.
```
MartinHjelmare, line 51:
```suggestion
  The user first needs to install a Linux-based operating system (i.e. Debian). The user then needs to run the installer script.
```

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,80 @@
...
+
+* **Start when the system is started:** done by the installer
+* **Run with full network access:** done by the installer
+* **Access USB devices:** done by the installer
+
+* **Maintenance Home Assistant installation**
```
MartinHjelmare, line 57:
```suggestion
* **Maintaining the Home Assistant installation**
```

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,80 @@
...
+  Home Assistant can be maintained from the Supervisor. Includes rollback when the update fails.
```
MartinHjelmare, line 58:
```suggestion
  Home Assistant can be maintained from the Supervisor. This includes a rollback when the update fails.
```

adr/0014-home-assistant-supervised.md
```diff
@@ -0,0 +1,80 @@
...
+
+  - **Python upgrades:** Included in Home Assistant updates
+  - **Installing Python dependencies:** Included in Home Assistant updates
+  - **Updating Home Assistant:** Via the UI
+
+- **Maintenance Operating System**
```
MartinHjelmare, line 64:
```suggestion
- **Maintaining the Operating System**
```

JumpMaster:
> @JumpMaster well, with the above, Ubuntu won’t be supported... 🤷‍♂️
> 
> Docker.io has caused many surprises in the past. As this is about what is supported, we of course will follow Docker, not Ubuntu’s maintained version of it.

@frenck Thanks for replying.  My thoughts are still valid regarding Debian.  For example Debian 10 would install docker.io 18.09 where as 19+ would need to be pulled from Docker directly.  I guess my question is whether it's better to support the version of Docker from the OS or from the creators themselves?

frenck:
It this case, the creator... as it is not just Debian to take into account, we have an Home Assistant Operating System to support as well.

borpin:
If Debian is the only supported distribution (and I think it should be made **crystal clear** it is vanilla Debian and not any Debian derivatives), Supervised is not going to be supported on an RPi. Is that the intention?

The line "The user first needs to install a Linux-based operating system (i.e. Debian)." is open to interpretation and in this context and should simply be "The user first needs to install Debian." (with a link to the Debian distribution page).

Also, you should include the requirement that apparmour is included in the **kernel**. It is possible to install the package but for apparmour to not be enabled as the installed kernel does not support it (as per the previous version of Raspbian - not sure if that has changed).

Jc2k:
@JumpMaster 

> Why Docker CE over Docker IO? On Ubuntu I've stuck to Docker.io as to get a fully supported product from Ubuntu.
> 
> My thoughts are based on the reply "Be wary of docker-ce" on this topic.
> https://stackoverflow.com/questions/45023363/what-is-docker-io-in-relation-to-docker-ce-and-docker-ee

Hopefully not off topic but wanted to note that the stackoverflow reply is oversimplifying things a bit, imho. See the debian packaging guide [here](https://go-team.pages.debian.net/packaging.html). Go libraries are generally packaged as source code, not binaries, which complicates some of the normal benefits to distro packaging. It won't mean you can update a go dependency like you can libssl. Go binaries still have all those libraries compiled in. What debian packaging does is make your GOPATH reproducible so you can rebuild the package with only an apt mirror even if the upstream git repos go away. This is a useful property but doesn't help with memory footprint like that s/o post suggests. This is still a binary and if its maintainer goes away you stop getting updates. So ultimately you end up gambling on which maintainer will give up first - Docker itself, or a debian volunteer... And the one who packaged docker-io in the first place already gave up and wasnt replaced for a year so...

JumpMaster:
@Jc2k 
Thank you very much for your detailed reply.  It was really helpful.

frenck:
@borpin it clearly states Debian 10, buster?

Raspberry Pi is supported by Debian, they provider images for the Pi.

borpin:
@frenck 
> it clearly states Debian 10, buster?
It does, but the line further down adds a little uncertainty. Perhaps add in 'official' - trying to be watertight for you 😁.

> Raspberry Pi is supported by Debian, they provide images for the Pi.
Never found that - excellent I'll give it a spin 👍

Not for Pi4 as yet though 😢 - bound to come eventually.

adr/0014-home-assistant-supervised.md
foxy82: If the only supported version is Debian then shouldn't this read "The user first needs to install Debian 10"

borpin:
> you should include the requirement that apparmour is included in the **kernel**. It is possible to install the package but for apparmour to not be enabled as the installed kernel does not support it (as per the previous version of Raspbian - not sure if that has changed).

Not sure if this suggestion was picked up.

balloob:
@borpin oops, missed them. Addressed now.

Kanga-Who:
`" No additional software, outside of the Home Assistant ecosystem, is installed."`

Does this statement also cover software installed using Docker? ie. If I install Portainer outside of HA, even thought it is a supported component.

finity69x2:
By adding the following you are completely negating the entire reason that the Supervised version of HA was fought to be continued:

```
This installation method can easily be broken if one manages the operating system incorrectly. Therefore, the following additional conditions apply:

- The operating system is dedicated to running Home Assistant Supervised.
.....
- No additional software, outside of the Home Assistant ecosystem, is installed.
```

I get that you don't want to be required to provide support for the host OS not being properly configured but i think that stance is bit draconian.

Does that mean that just because I have installed any other non-HA related package/container **at all** that you will no longer provide support for HA even if the installed package/container has no impact on the functioning of HA or docker?

balloob:
Home Assistant Supervised is not an application that you install on your system, the system becomes Home Assistant Supervised. The Supervisor expects to be able to control every aspect of the system.

If we are to say that you can run any software you like, and that software doesn't work correctly with Home Assistant Supervised, it would be up to us to fix. 

Portainer is a great example of software that can really mess with the system. It allows you to really break your Home Assistant installation, because you can change Docker containers that the Supervisor expects to manage.

Kanga-Who:
@balloob `"Portainer is a great example of software that can really mess with the system. It allows you to really break your Home Assistant installation"`

In the wrong hands, sure. I was only using Portatiner as an example. 

How about an MQTT server? If I am to run an MQTT server on the same machine and after an HA update the machine won't boot, the response would be, "sorry - no support, you broke your own system by using something else", even if the issue with HA had absolutely nothing to do with the MQTT server, or its dependencies?

balloob:
> In the wrong hands, sure.

This is exactly the problem 🙂 

These ADRs are all about expectation management. You do it in one of our four prescribed ways and we can tell you exactly what expertise and maintenance is required from you and what you can expect Home Assistant to take care off.

You can run Debian 10 with Samba and SSH. It probably will be fine. But if it isn't and it breaks with an update, it should be clear that you're going to have to fix this yourself. Maybe there will be a community guide that can help you or you can get help on the forums.

We cannot test and guard for every possible permutation of installed apps in an OS with a system that expects full control. 

And if Home Assistant Supervised is too invasive, Debian + Portainer + Home Assistant Container is available as an alternative.

Kanga-Who:
> We cannot test and guard for every possible permutation of installed apps

I understand this and agree. So limit those supported to key programs.

If the base OS is being locked at support for Debian only, which I have no issue with, then a limited set of programs known to not effect HA operation could also be supported as a minimum. SSH could be one of these.

I have been an avid user of a Supervised install until very recently and strongly advocated for it's continued support. I have _always_ run SSH and Samba, amongst many others, at an OS level and have _never_ had an issue with HA not working due to these being installed.

frenck:
> No additional software, outside of the Home Assistant ecosystem, is installed.

@Kanga-Who Yes, that includes additional containers outside of the ecosystem.

frenck:
> If the base OS is being locked at support for Debian only, which I have no issue with, then a limited set of programs known to not effect HA operation could also be supported as a minimum. SSH could be one of these.

Those are not key programs. As they are all are available as add-ons, and thus available within the supported ecosystem.

Kanga-Who:
> This is actually an incorrect statement; Add-ons run independently and are available even when e.g., a Core update fails.

Not in my experience, which was the entire reason I switched from using those particular integrations. Perhaps that has changed in more recent time.

frenck:
I'm sorry @Kanga-Who, it has always been that way. Other Docker containers (add-ons) are not affected and keep running.

Kanga-Who:
> I'm sorry @Kanga-Who, it has always been that way. Other Docker containers (add-ons) are not affected and keep running.

Like I said, in experience, that was not the case. An HA update failed, and the HA system was completely unavailable and unresponsive using the built in add-ons. This is getting off topic, so I will leave it at that if you are assuring me that this would not be the normal situation.

I'm just pleased that Supervised is continuing to be supported 👍

magnusoverli:
I have been using the Supervised install method for a couple of years now. For me this has been a pleasant journey of learning both HA and Ubuntu. I recently switched over to ESXi so I am personally off of the generic install train.

But, I fell that HA need to cater to the user group that doesn’t have the hardware needed to run a supported HassOS installation or the skills needed to set up and manage a hypervisor. For everything that is wrong with Ubuntu (in a HA-perspective) it is also the Linux distro that has the most appeal to beginners, and the documentation and tutorials available online is insane! 

When the decision to deprecate the generic install method was published, I was at first scared for what it meant for me and my HA usage in the future. But I also supported the idea, since I know that the core team, the very helpful users and the community in general spends an insane amount of time tracking bugs and issues that stems from outside the HA environment, wether or not is initiated by the end user.

For the reasons mentioned above, I feel that if a generic install is going to be supported in the future, Ubuntu has to be a part of that. Otherwise I think that we are better off deprecating the install method all together...

Best regards...
MagnusO

balloob:
Part of this ADR describes the required expertise necessary for the supervised installation. Do you agree with that description?

Beginners do not have the required expertise. The goal is very much to guide beginners to other installation methods that require little maintenance.

Tutorials are great for people to get started with things. But installation is just a small step, after that you end up with a system that needs to be maintained. You need to maintain the right version of components as the HA Supervised installation needs them, which might not match what is available in the package manager.

magnusoverli:
Hi @balloob!

I do agree on the required expertise and responsibilities described. No issue with that!
My concern is that by taking this step, it becomes harder for the new user to take his/her first step into the HA domain without investing in hardware. Given, the minimum required investment isn’t much for a RPi3, but still. 

I can’t see anything good come of this as it stands now. The work load will probably go down a little, but not as much as you hoped when the original decision was made.  At the same time there are strict requirements in place (rightly so) that keeps the up and coming tweakers from getting started. I also see a future in which people with this described generic installation will be taken apart in the forums and chat rooms, simply because we will assume that the issue is originating from the user. It might very well be, but it will provide a mind-set, enabling more hostile interactions.

The generic install method will eventually die after a while, and that’s fine, sort of. This just feels like a pointless middle ground option. 

My 2 cents are simply this; all or nothing! :-) 

☺️

nickrout:
> I'm sorry @Kanga-Who, it has always been that way. Other Docker containers (add-ons) are not affected and keep running.

@frenck one question, do addons run if supervisor fails? (which happened to me the other day, I think during a supervisor update). I am running supervised on debian 10 (64 bit) and I think that other than the inclusion of the OS level ssh it complies with this ADR.

balloob:
> My concern is that by taking this step, it becomes harder for the new user to take his/her first step into the HA domain without investing in hardware.

This is an interesting point. What makes this the more appealing or easier approach compared to a VM or Docker container? The VM needs just to be imported and the container can be started with a single Docker command.

Or is it the add-ons and update via the UI? That's exactly what is made possible by having the Supervisor manage the system, which means we can't have the system change under us.

I think that the goal of an all-in-one system for beginners that can run on Ubuntu cannot be achieved with the approach that this installation method takes. Maybe it would be better if it is implemented as a Snap that can be updated via the Snap store?

The Supervised install is for experts and I don't think that it will eventually die out. If a user prefers a different platform, they can follow one of the Supervised community guides ([thanks Kanga!](https://community.home-assistant.io/tag/installation)) instead of a guide on home-assistant.io to get started. The guide can start including common problems and how to solve it for the platform the guide covers.

amelchio:
I think that I agree with @magnusoverli; he says this "feels like a pointless middle ground" and I said above "I don't really see the use case".

Because if you are expert enough to maintain this installation method, you are going to want to amend the supported setup and no recommendation will hold you back. Agree? Or do you see a kind of user that I am missing?

Maybe it is not really expected that anyone will run a "pure" Supervised install as supported here? The installation method could be seen more as a common base for community guides to build on?

In that case, the description in this PR (and even the installation guide) could be phrased to clearly state that this is a building part, not a finished product. Using anything built on this will be community supported.

It could even be called Unsupervised as in "you are on your own" (that was a joke).

foxy82:
> This is an interesting point. What makes this the more appealing or easier approach compared to a VM or Docker container? The VM needs just to be imported and the container can be started with a single Docker command.

I think the VM is far from "just to be imported" you need to make sure virtualisation is enabled in your BIOS, you need to create bridged network adaptors, which if you don't understand networking can be hard as well as losing any static IP set in your router as the mac address changes. You then need to make sure you pass all your usb/serial devices through to the VM as well. Plus ther could be a whole host of VM hosts running on different OSes.

ties:
> Because if you are expert enough to maintain this installation method, you are going to want to amend the supported setup and no recommendation will hold you back. Agree? Or do you see a kind of user that I am missing?
> 
> Maybe it is not really expected that anyone will run a "pure" Supervised install as supported here? The installation method could be seen more as a common base for community guides to build on?
> 
> In that case, the description in this PR (and even the installation guide) could be phrased to clearly state that this is a building part, not a finished product. Using anything built on this will be community supported.

A project I maintain has a decent number of support tickets filed (and likely even more users hitting issues) caused by outdated installation guides that were made by third parties. For the supervised install I imagine the same thing will be the case if the installation method is provided.

The only reasonable way out of that situation I see is that a user has to actively acknowledge that they are running an non-supported setup (if you can detect it at all). For the supervised install in my eyes it would be best if this warning was part of the onboard ng process as well as the instructions.

magnusoverli:
> > My concern is that by taking this step, it becomes harder for the new user to take his/her first step into the HA domain without investing in hardware.
> 
> This is an interesting point. What makes this the more appealing or easier approach compared to a VM or Docker container? The VM needs just to be imported and the container can be started with a single Docker command.

If I remember correctly from my own experiences, the appeal here is mainly the add-on environment and the UI updates. What the add-ons offer to newcomers and beginners is hugely important! The install method in itself is not important in itself, other than VM´s and Docker is going to be a VERY steep curve to ask of the beginners. If we cannot offer an easy, supported and free way into a supervised HA environment, even with limited performance/lifespan, I believe were gonna miss out on a future generation of home automators and tweakers. 

I do not have the answers here, and I know I am currently nit-picking and playing the devils advocate, but I feel the argument is valid. :-)

balloob:
@amelchio I think that you're right that we can do better in phrasing here about the expected audience and how to use it. 

What about something like:

This installation type is meant for experts. It delivers an integrated system that runs Home Assistant and its Supervisor. If you choose to customize your Debian installation by installing other software, Home Assistant no longer guarantee that it keeps working.

magnusoverli:
A question though... Is there a different intended audience between this method and a VM/Docker setup?

borpin:
> This installation type is meant for experts.

Instead of 'experts' how about 'users with a good working knowledge of Linux'

> Home Assistant no longer guarantee that it keeps working.

I don't think you want to 'guarantee' that. - more that 'the development team will be unlikely to be able to provide support in the event of conflicts.'

scstraus:
>That does not mean it will not work on Ubuntu which is Debian based. Ubuntu in it's state today could be considered as compatible. But if Ubuntu 21 removes systemd or AppArmor it will become incompatible. Home Assistant will make no changes to support Ubuntu. If the next release of Debian removes systemd or AppArmor Home Assistant will be changed to support it.

Very happy to see supervised going forward on generic linux, even with some limitations. I can definitely live with them. 

As someone who currently runs supervised on Ubuntu 18.04 LTS with docker-CE with almost zero change to vanilla OS (a common install from what I’ve discussed with others) it would be very helpful to be given an “end of life” date for such an install. I have no problem moving to Debian, but I would prefer to not have to do it right away if at all possible. 

Would be nice if a bit of leeway would be given so that we can migrate to the supported architectures

KennethLavrsen:
I just read the document

I seems you guys do not understand why we are many that want the Supervised installation method.

We want to run other services on the same machine! It becomes totally poinless running supervised if you cannot install anything else. I may as well install your operating system.

I run Supervised because I run a couple of other non-Docker services on the same machine and because I want to run a proper OS with a large community like Debian or Ubuntu to ensure the system is safe.

I am OK that the OS must be the Debian family and I am OK with the requirement that the Docker must be the specific flavour. I am also OK with Home Assistant being the only Docker based installation. And expecting a current stable flavour of Debian or current Ubuntu LTS is also a reasonable requirement.

But don’t expect people to not install other standard daemons using apt install on the same machine.

I will want run ssh, and Samba natively. It is pointless to run these as addons. And I want to run other simple clean apt package installed services that I use as well. I have other Gods than Home Assistant and my home has plenty of stuff that may be monitored and controllable by HA but still running just fine, even when HA and its containers have stopped or crashed or being upgraded. There are so many breaking changes and stability issues with both HA and addons that we are many that want the things we really rely on to work when HA is not there.

That is why we run Supervised!

Jc2k:
AIUI this is about defining what the developers support with their limited resources, not about actively blocking anything. There is no getting away from the fact that supporting this package leads to a reality where the team is going to get support requests were users with less experience than you have samba half configured on the host and then try to install the samba addin. Those kinds of support requests are going to take a lot of time and energy from the team. So I fully support that they want to try and define a clear boundary between the service they are committing to providing and the space where the community can come together and shine. Especially as the vast majority of the HA contributors are unpaid volunteers.

Which bits of supervised do you use? It sounds like an expert user like yourself would be better off with the core container and no supervised? I’ve never run supervised myself, i love the idea of Home Assistant OS for non techie users especially but I’m happier managing the OS myself too. What am I missing out on?

BTW I really like Home Assistant OS from a security POV- minimal buildroot OSs have a much smaller attack surface. I have loads of patches on Ubuntu for packages that aren’t even installed on a typical buildroot based image.

balloob:
@KennethLavrsen did you read the comments?

KennethLavrsen:
I love two features of the supervisor part of Home Assistant

1. The very user friendly one click upgrade of Home Assistant. You have the new version announced when available and it is an easy upgrade.
2. The snapshot feature. I can snapshot both my configuration and I can easily roll back to a previous fully working combination of HA revision and configuration. I use the snapshot feature several times per week.

The addons I use are all HA releared. I use the good old File Editor as it works very well on an iPad when I am not in front of a computer. And I use the Visual Studio Code editor which is great on a larger monitor and real keyboard. Then I use the community ssh addon to access HA from within the HA docker environment. I rarely use this though. And I use the SQLite web to study the data logged. These are all features that would be hard to setup and are directly related to HA itself. The  supervised installation is a great thing.

What I do not want is to run daemons that are not part of HA as HA addons. This is where the supervised installation on a full Linux OS fits perfectly.

balloob:
I have updated the phrasing based on the feedback. I have now rephrased the support sentence in a way that readers can see the reasoning behind it without having to read all the comments in this thread:

> Home Assistant is not responsible for issues that come up with your host or config/resource conflicts caused by other applications installed on the host system.

amelchio:
It seems like a bit of weasel wording to mention that Home Assistant is not responsible for something. It implies that Home Assistant _is_ responsible when the installation procedure is followed but that is not quite true; there is no SLA, everything is never more than "best effort".

I think it would clear things up to acknowledge that this installation method on its own will be useful to almost nobody. That it is fine to modify it, we are just not going to help you with that. Maybe something like this:

> Home Assistant Supervised is provided as-is as a foundation for community supported do-it-yourself solutions. We only accept bug reports for issues that have been reproduced on a freshly installed, fully updated Debian with no additional packages.

JumpMaster:
I think you're missing the point. Home Assistant will be built and tested to work on Debian 10 in a supervised configuration. When Debian 11 is released Home Assistant will support it in a pre-acknowledged timeframe. A new feature of upgrade of Home Assistant will never knowingly break the supervised installation.

If you then as an advanced user install KVM for example which breaks the networking stack.  You can't raise a GitHub issue as isn't a supported / tested / guaranteed configuration.

We all want the devs to be developing and not just supporting the endless configuration scenarios.

pvizeli:
@JumpMaster the first person which brings the problem to the point. Thanks

jsiemek:
Let me add my point of view.
For over 2 years I run my supervised install on Ubuntu 18.04.4 with tons of other software on it. I will not migrate to Debian and not dedicate i5 with 16Gb RAM to HA only. 
Even I have spare RPi3 I will as well not migrate to it, as for example Visual Studio Code add-on will not work on it. As for former Oracle employee running DB on SD card is terrible.
But I’d like to stay on supervised version for it’s usability. 
I’ll take care of my install on my own, no problem.
So simply display BIG WARNING that such install is not supported and let my and many similar go on with installation.

KennethLavrsen:
I have read all comments and also the latest updates.
And I still claim that you have not understood WHY people will want Supervised installation method.

You write in the document

"We only accept bug reports for issues that have been reproduced on a freshly installed, fully updated Debian with no additional packages."

This is useless. Why would anyone have a supervised installation if they are not allowed to install extra packages unconditionally? What is the point of the method if I cannot install a couple packages like the ssh daemon or maybe the Deconz application on bare metal like I run? It is not reasonable to be that harsh in this statement. This should be softened up.

I can understand that if a daemon causes a conflict then it is my own responsibility to fix the conflict. And there may be packages that alter the system in a way that they will never be compatible. I can understand that you do not want to spend time on sys admin support. But I should still be able to raise a bug report for HA itself if a timer feature does not work or the Google integration has a bug. I should not have to lie in the bug report that I have not installed anything on my server.

These two bullets also make the supervised installation pointless

"The operating system is dedicated to running Home Assistant Supervised."

"No additional software, outside of the Home Assistant ecosystem, is installed."

These two statements make the whole idea absurd. If the condition is that you cannot install ANYTHING on the server then there is no point in making such an installation in the first place.

Then we might as well just install the Home Assistant OS version and accept that you just own exclusive rights to deciding what is installed on my machine.

It would be more reasonable to write

"The operating system has all dependecies for Home Assistant Supervised un-altered and in the latest official up to date revisions"

"You are responsible to ensure that any software you install, which is outside the Home Assistant ecosystem, does not creates conflicts that prevents Home Assistant from working. If you install addons that conflicts with services you have installed on the OS then it is your own responsibility to resolve these conflicts yourself or remove the conflicting service."

The purpose of a Supervised installation is to enable running Home Assistant on a machine that has another purpose. It could be as a mail server. It could be as a file server with a large harddisk and Samba share for general storage. These will never conflict with Home Assistant unless you try and install an addon that does the same thing using the same resources.

frenck:
> The purpose of a Supervised installation is to enable running Home Assistant on a machine that has another purpose.

Sounds like your use case? Running a Hypervisor is another option for example, which perfectly allows for that as well.

> It could be as a mail server. It could be as a file server with a large harddisk and Samba share for general storage.

Sure, you can do that in another VM or on the same machine, but the latter will not be officially supported. It is not disallowed; it is your machine.

> These will never conflict with Home Assistant unless you try and install an addon that does the same thing using the same resources.

**There we go!** Your last sentence hits the pain point with this installation method straight up! 😁

"If", you might understand that, unfortunately, a lot of people don't. Which is part of the issue of why this whole thing was fired up.

In your example: If the add-on doesn't install or start, who is the one to blame? Unfortunately, most raise an issue with Home Assistant or the add-on maintainer because "it doesn't work".

Santobert:
Just to be sure: The supervised installation method results in almost the same user experience as the operating system method. It is not supported to change the underlying operating system in any way.
**But** this is allowed.

> The purpose of a Supervised installation is to enable running Home Assistant on a machine that has another purpose.

So the purpose of the supervised method is to allow the installation of additional packages, even though it is not officially supported.

I'm afraid that this will lead to confusion. Why is this installation method supported, but the only purpose of this installation method not? Just to prevent issues that are caused by the user?

JumpMaster:
> Just to prevent users from raising issues that are caused by themselves?

Bingo!  Yes that is the issue.  HA supervised on Debian 10 definitely works.  You have broken it.  You are responsible for fixing it.

thomasloven:
Please refer to the [ADR-0012 proposal](https://github.com/home-assistant/architecture/pull/392) for clarification of the meaning of "supported" in this context.

scstraus:
> > Just to prevent users from raising issues that are caused by themselves?
> 
> Bingo! Yes that is the issue. HA supervised on Debian 10 definitely works. You have broken it. You are responsible for fixing it.

As someone who runs supervised and who's a former Unix sysadmin, I just want to say thank you to all the people here who listened to the community and kept this form of install available to people like me. I am going to try to adhere to all the rules, but just having the guidelines of the correct way of doing this and dependencies should be enough for me even if I do stray off the supported path. 

The rules are there for a reason, to keep the people who need them in a place where they won't fall off a cliff. For those of us who know enough to consider breaking one of them to do something that isn't supported, we will have to figure out our own problems. 

I feel that this is by far the best compromise, and I know it comes at the cost of some cycles from the supervisor dev team, and I just want to say thanks to them for putting the effort in to give us the most options. There are a lot of reasons why other install options don't work for me and I just want you to know that it's really appreciated.

balloob:
Thanks everybody  for the input. We'll get to work on reinstating the installer and updating the documentation.

PR #395, opened 2020-06-08
 ADR: Home Assistant OS as supported installation method

balloob:
This method was previously known as "Home Assistant". Feedback said this was causing confusion so defining this installation method should also rename it. I put in HA One but if someone thinks of a better name, let's discuss.

### DIFF
--- a/adr/0015-home-assistant-one.md
+++ b/adr/0015-home-assistant-one.md
@@ -1,0 +1,82 @@
+# 0015. Installation method: Home Assistant One
+
+Date: 2020-06-08
+
+## Status
+
+Accepted
+
+## Context
+
+Define a supported installation method as per [ADR-0012](https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md).
+
+## Decision
+
+Home Assistant One is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
+
+This is the general recommended installation method and the one that our website and documentation should be focused on.
+
+### Supported boards/hardware/machines
+
+- Raspberry Pi 2
+- Raspberry Pi 3 Model B and B+ 32bit
+- Raspberry Pi 3 Model B and B+ 64bit
+- Raspberry Pi 4 Model B 32bit
+- Raspberry Pi 4 Model B 64bit
+- Tinkerboard
+- ODROID-C2
+- ODROID-N2
+- ODROID-XU4
+- Intel NUC
+- Virtual Machine (64 bits intel-based)
+
+### Supported Operating Systems and versions
+
+This installation method is only available and supported using the Home Assistant Operating System. Only the latest major version is supported (Currently, 4.x).
+
+When a new major version is released, the previous major version will be dropped with a deprecation period of 2 months. The last 3 minor releases are supported.
+
+### Supported Hypervisors
+
+The Home Assistant Operating System can be run on a Hypervisor and thus be run as a virtual machine. The following Hypervisors are supported:
+
+- TBD. Maybe:
+- VirtualBox
+- Proxmox
+- UnRaid
+- VMWare ESX
+- VMWare Workstation
+- HyperV
+
+### Required Expertise
+
+- **Installation**
+  Installation requires the user to install the disk image for their device, install an image flasher (Etcher) and flash the image to the disk. Then insert image and boot their device.
+
+* **Start when the system is started:** Managed by installer
+* **Run with full network access:** Managed by installer
+* **Access USB devices:** Managed by supervisor
+
+* **Maintaining the Home Assistant installation**
+  Ongoing maintenance can be done via the user interface. User will be notified if an update is available for the operating system or for Home Assistant. Updates will be automatically rolled back if they fail.
+
+  - **Python upgrades:** Managed via HA updates
+  - **Installing Python dependencies:** Managed via HA updates
+  - **Updating Home Assistant:** Via a UI
+
+- **Maintaining the Operating System**
+  Ongoing maintenance can be done via the user interface. User will be notified if an update is available for the operating system. Updates will be automatically rolled back if they fail.
+
+* **Security updates for OS:** Managed via HA OS updates
+
+* **Maintaining the components required for the Supervisor:** Managed via HA OS updates
+
+**Conclusion:** low expertise required. Besides the installation methods, everything is done via the UI.
+
+## Consequences
+
+Update documentation on how to install this method, the required experience and the expected maintenance.
+
+Move existing related documentation that does not match the requirements to the community guides wiki.
+
+Add checks to Home Assistant to warn (but not forbid) when not all requirements are met for this installation method.
--- a/adr/0015-home-assistant-one.md
+++ b/adr/0015-home-assistant-one.md
@@ -1,4 +1,4 @@
-# 0015. Installation method: Home Assistant One
+# 0015. Installation method: Home Assistant All-in-One
 
 Date: 2020-06-08
 
@@ -12,7 +12,7 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 ## Decision
 
-Home Assistant One is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
+Home Assistant All-in-One is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
 
 This is the generally recommended installation method and the one that our website and documentation should be focused on.
 
--- a/adr/0015-home-assistant-one.md
+++ b/adr/0015-home-assistant-one.md
@@ -1,4 +1,4 @@
-# 0015. Installation method: Home Assistant All-in-One
+# 0015. Installation method: Home Assistant OS
 
 Date: 2020-06-08
 
@@ -12,7 +12,7 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 ## Decision
 
-Home Assistant All-in-One is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
+Home Assistant OS is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
 
 This is the generally recommended installation method and the one that our website and documentation should be focused on.
 
--- a/adr/0015-home-assistant-one.md
+++ b/adr/0015-home-assistant-one.md
@@ -14,11 +14,10 @@ Define a supported installation method as per [ADR-0012](https://github.com/home
 
 Home Assistant One is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
 
-This is the general recommended installation method and the one that our website and documentation should be focused on.
+This is the generally recommended installation method and the one that our website and documentation should be focused on.
 
 ### Supported boards/hardware/machines
 
-- Raspberry Pi 2
 - Raspberry Pi 3 Model B and B+ 32bit
 - Raspberry Pi 3 Model B and B+ 64bit
 - Raspberry Pi 4 Model B 32bit
@@ -40,13 +39,12 @@ When a new major version is released, the previous major version will be dropped
 
 The Home Assistant Operating System can be run on a Hypervisor and thus be run as a virtual machine. The following Hypervisors are supported:
 
-- TBD. Maybe:
 - VirtualBox
 - Proxmox
 - UnRaid
-- VMWare ESX
-- VMWare Workstation
-- HyperV
+- VMWare
+- Xen
+- Qemu
 
 ### Required Expertise
 
--- a/adr/0015-home-assistant-one.md
+++ b/adr/0015-home-assistant-one.md
@@ -40,11 +40,16 @@ When a new major version is released, the previous major version will be dropped
 The Home Assistant Operating System can be run on a Hypervisor and thus be run as a virtual machine. The following Hypervisors are supported:
 
 - VirtualBox
-- Proxmox
-- UnRaid
 - VMWare
 - Xen
 - Qemu
+
+We will provide documentation for the following systems build on top of these technologies:
+
+- VirtualBox
+- VMWare
+- Proxmox
+- UnRaid
 
 ### Required Expertise
 


adr/0015-home-assistant-one.md
```diff
@@ -0,0 +1,82 @@
...
+
+The Home Assistant Operating System can be run on a Hypervisor and thus be run as a virtual machine. The following Hypervisors are supported:
+
+- TBD. Maybe:
+- VirtualBox
+- Proxmox
+- UnRaid
+- VMWare ESX
+- VMWare Workstation
+- HyperV
```
ludeeus, line 49: I know the list is TBD, but Hyper-V does not have USB passthrough, due to that there are a few things that will not work.
I do not feel that it should be a part of supported hypervisors for that reason alone.

frenck: Good point, agree on that.

pvizeli: Let's rename it to Xen

pvizeli, line 49:
```suggestion
- VirtualBox
- Proxmox
- UnRaid
- VMWare
- Xen
- Qemu
```

adr/0015-home-assistant-one.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Decision
+
+Home Assistant One is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
+
+This is the general recommended installation method and the one that our website and documentation should be focused on.
+
+### Supported boards/hardware/machines
+
+- Raspberry Pi 2
```
DubhAd, line 21: Currently the docs say _not recommended_ - if we're going to say that there, we should say it here.

adr/0015-home-assistant-one.md
frenck: Should we document bare metal Xen & Qemu?

adr/0015-home-assistant-one.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Context
+
+Define a supported installation method as per [ADR-0012](https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md).
+
+## Decision
+
+Home Assistant One is the full installation of our all-inclusive home automation system. Best in class home automation is complemented with a UI for configuring your system, making backups and safe updates with automatic rollback.
+
+This is the general recommended installation method and the one that our website and documentation should be focused on.
```
DubhAd, line 17: This is the _generally_ recommended installation

dgomes:
Shouldn't the installation methods be self-explainable ? 

Home Assistant One -> Home Assistant Image 
Home Assistant Container -> Home Assistant Container
Home Assistant Supervised -> Home Assistant Install Script ?

frenck:
Some background information on "Home Assistant One".

The "One" was put here up as a suggestion because it is the first installation method we want to push. "The one", "The one in all", "The one with all the project offers", "The one that brings the full experience", "The one to go to when starting out", "The one that makes it easier".

Some other suggestions, including my personal feeling with it:

- **Home Assistant Complete**
  To me sounds like a vitamin pill 😂  But that is mainly it, to be honest...
- **Home Assistant OS (or Operating System)**
  Ambiguous, as we actually have an [Operating System](https://github.com/home-assistant/operating-system) project, which is just part of the thing, not the whole thing.
- **Home Assistant Image** (as suggested by @dgomes just now)
  Makes sense in a way...

The other suggestion for Home Assistant Supervised to Install Script is something to avoid. Maybe it becomes a package later? The install script won't be the part that one is running eventually... 🤷

dgomes:
The ADR says *installation method*, I was not debating the "One" (no special feelings pro/against the name)

So my proposal is to refer to the *install methods* as:

- Home Assistant One Image 
- Home Assistant Container
- Home Assistant One "Install Script" / "wizard" / "guided"

frenck:
Aah right, but in that case we would end up with two methods that have "One": `Home Assistant One Image` and `Home Assistant One "Install Script" / "wizard" / "guided"`

It becomes ambiguous really quick if someone says: "I run Home Assistant One", well which ones? Which is a problem we experienced with "Home Assistant Core" as well. (Python or Docker?)

Hence the current naming is making sure it sounds and look different to avoid that as a whole.

dgomes:
My issue, is that what you say implies you have two different experiences *after install*.

SeanPM5:
I like Home Assistant One personally. Not a huge fan of "Complete" because of what Frenck mentioned; plus when it's abbreviated it will be too close to HACS and cause confusion imo.

Naming stuff is hard, only other thing I can think of is maybe "Home Assistant Max" ?

frenck:
@dgomes You actually have 2 different experiences after install. The biggest would be the difference in automated updates to the host operating system (with automatic fallback on upgrades). While the other has quite a level of experience needed to maintain it yourself. The bare fundament it runs on is different.

dgomes:
Call it:

- Home Assistant **Two** "Install Script" / "wizard" / "guided"

My point is still the same, the *install method name* should be self explanatory in what is involved

DavidFW1960:
What is the difference between Home Assistant One and Home Assistant Supervised?
Is One including HassOS and Supervised is the old generic Linux install?
Home Assistant Zero?

jfmcarreira:
I think it would make sense to call it "Home Assistant Image" or "Home Assistant OS". 

@frenck mentioned that HassOS is a different project, but in the end it is just ground where Home Assistant can be run. One will only have HassOS for Home Assistant. Or is there any other purpose?

And I am thinking of the user perspective when looking for tutorials/support. He just says, "I am running Home Assistant OS" exactly what it is.

Moreover, and sorry for being off-topic, I think the OS updates should be paired with HA updates, in the sense that if one is running HA version 0.110 it should always be in a specific OS version (at least for the stable releases). This way everything is tight together and support would be much easier I think.

tdejneka:
In 1956 when the Ford Motor Company was developing a new car, they needed a name. They received thousands of suggestions and even reached out to poet Marianne Moore who produced truly inventive ones:

Utopian Turtletop
Andante Con Moto
Mongoose Civique
Pastelogram
Intelligent Bullet
Bullet Cloisonne

Ultimately, the company chose to use Henry Ford's son's first name: Edsel.

Based on the car's lack of success, never let it be said that any old name will do.

---

With just a bit Moore's whimsy, I give you, in descending order of product completeness/complexity, the four products:

1. Home Assistant Plus
2. Home Assistant Minus
3. Home Assistant Coretainer
4. Home Assistant Core

... and no, "Coretainer" isn't a typo.

If it's a toss-up between "One" and "Image", "One" has my vote.

adr/0015-home-assistant-one.md
pvizeli: I simplify the list and try to solve the question, it's now semi correct :)
I can say for sure, it works for Xen because that is what all releases are tested like Proxmox in a future step.

pvizeli:
Home Assistant OS is a beautiful name, but it already describes the project, and we should not do the same issue as with Hass.io with described at the end two different paths. We can not use that name or need to rename the OS, which is technically not possible without breaking all current installations.

Adminiuga:
Home Assistant All-in-One

shbatm:
My only thought about Home Assistant One is whenever you do get to a 1.0 release and beyond is that going to add another round of confusion?

i.e. Home Assistant Core 1.0 vs. Home Assistant One?

dshokouhi:
I personally think that Home Assistant OS is the way to go.  I completely understand from a development perspective why we are opposed but in all honesty from an end user perspective (which is really what this conversation is all about) its odd to have that name reserved for something else.  So think about when you install Windows for example.  With the Windows OS you get apps like paint and others bundled in with it, its not like your missing key features like a file explorer that is expected to be included.  So as a user when you install Windows you expect to see certain apps and features.  In contrast Home Assistant OS would do exactly the same thing.  You not only get a running system with BIOS and containers and what not but you get Home Assistant along with it, which is the users end goal.  To a user the HA OS would be a complete package.  They install the HA OS with the intention of Home Assistant.

With that said I like Home Assistant All In One as an alternative.  You can also call it Home Assistant Bundled for that case.  

Also why not put this in the hands of the users? Put out a survey with all the suggestions we get here and see what makes sense to the actual user base.

balloob:
I am starting to like Home Assistant OS too. It's actually flashing the HassOS images that give you that what is called Home Assistant One :) We wouldn't have to rename any of the existing operating-system repo's etc. I think those namings are all fine as we're only naming the installation method here.

tdejneka:
It's important not to ignore the lesson learned from the previous rebranding initiative.

Changing `hass.io` to `Home Assistant` caused confusion (and resistance) because both terms had co-existed for years and had a specific meaning understood by users and recorded in documentation and external resource materials. Co-opting an existing name could not erase years of usage.

Changing `Home Assistant` (formerly known as `hass.io`) to `Home Assistant OS` would be a repetition of the same mistake. `Home Assistant OS` is already understood to be the custom  operating system included with the image-based releases for small-board computers. It has an established meaning to users. 

Surely we are capable of inventing a fresh new appealing name that doesn't simply recycle an existing one.

jfmcarreira:
The OS isn't name HassOS? I think this is the name users use. And this name should not be re used I agree.

But Home Assistant OS (HAOS for example) would be new.

I also enjoy names like All-in-one as opose to just One. All-in-One says it all :)

Finally I agree that re using names would create huge confusion for new users.

Adminiuga:
Instead of "Home Assistant OS" could you call it just "Home Assistant System"? After all it is a bit more than just "operating system".

shbatm:
I don't mind the Home Assistant OS name either, if it's not going to require a lot of retroactive renaming and @adminiuga has a point with 'System'.

Tossing some others out:

- Home Assistant IS (integrated stack) or FS (full-stack) 
- Home Assistant Prime
- Home Assistant Central
- Home Assistant Suite

tdejneka:
> Home Assistant OS (HAOS for example) would be new

The name `Home Assistant Operating System` is already in use.

Screenshot from 2020-06-09 19-18-41

In the community forum the name is abbreviated to `Home Assistant OS`.
https://community.home-assistant.io/search?q=Home%20Assistant%20OS

Recycling that name to now mean something else is the same mistake made in January when `hass.io` was renamed to something already in common use (`Home Assistant`).

`Home Assistant Suite` works for me. It's a brand new name and implies a collection of things (which is what `hass.io` has always been). Alternately, `Home Assistant Prime` is a riff on `Home Assistant One` and sounds loftier.

finity69x2:
I was pretty hooked on "Home Assistant OS" but seeing the other suggestions here I think I like "Home Assistant Prime" the best. Or even "Home Assistant Complete".

All three of those describe the end result much better than "Home Assistant One".

balloob:
I was talking with @cogneato and he came up with the suggestion Home Assistant Hub. I like that a lot, because when I introduce Home Assistant to people, that's what I say: it's an open source home automation hub.

pvizeli:
Yeah `Home Assistant Hub` looks fine too

jfmcarreira:
Home Assistant Hub sounds very interesting. There are other projects like Nextcloud which also use the word "Hub" to describe the combination of different tools in the same place.

nickrth21:
Short names are easy to remember, so "One", and "Core" are good, but "Container" and "Supervised" are a bit long. Perhaps:

Full OS -  **One** - Short, Simple and the place to start.
Supervised - **Moot** -  A meeting, gathering
Container - **Étui** - a case, container
Python App - **Core** - Already in use and represents that it doesn't have all the extras

The fun bit is that they rhyme with one two three four!

Frits1980:
Just my two cents:
Home Assistant key (like a cloud key) = hass.io
Home Assistant core = just core
Home Assistant sup = HA supervised for Linux
Home Assistant vm = venv

I don't think we should make it cryptic like "one" or "zero" or anything alike, because that would make it harder to understand for new people.

adr/0015-home-assistant-one.md
```diff
@@ -0,0 +1,82 @@
...
+- ODROID-N2
+- ODROID-XU4
+- Intel NUC
+- Virtual Machine (64 bits intel-based)
+
+### Supported Operating Systems and versions
+
+This installation method is only available and supported using the Home Assistant Operating System. Only the latest major version is supported (Currently, 4.x).
+
+When a new major version is released, the previous major version will be dropped with a deprecation period of 2 months. The last 3 minor releases are supported.
```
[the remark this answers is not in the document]
amelchio: I understand what 3 minor releases is but I do not understand what _supported_ means here.

Adminiuga:
> I don't think we should make it cryptic like "one" or "zero" or anything alike, because that would make it harder to understand for new people.
> Home Assistant sup = HA supervised for Linux

I'd argue "Home Assistant sup" is not less cryptic. It is all relative. New users just should read the ReadMe.md or "Where to start" and we  just  properly document it.  In other words, as long as it is documented, no name is too cryptic.

SeanPM5:
> I'm also in favor of opening this up to the community for ideas/voting.

Agreed, just posted a [forum thread](https://community.home-assistant.io/t/help-us-name-a-home-assistant-installation-method/204166) asking for ideas.

Kanga-Who:
I don't _completely_ dislike Hub, One or Plus, BUT, they are all used widely, or associated with other brands, like; 

- Smartthings Hub
- Disney Plus
- Pepsi One

Home Assistant should look for a unique name. I suggest **Home Assistant Genesis.**

tdejneka:
> look for a unique name. I suggest Home Assistant **Genesis**.

Hyundai and Weber would like a word with you. ;)

https://en.wikipedia.org/wiki/Hyundai_Genesis
https://www.weber.com/CA/en/grills/gas-barbecues/genesis-ii-series/

Kanga-Who:
> Hyundai and Weber would like a word with you. ;)

Point taken.

In all fairness, I've never heard of a Weber Genesis and the Genesis car brand is almost completely non existent in Australia, so that didn't even come to my mind.

Still a good name though :smile:

Kanga-Who:
**Home Assistant Fuse**
"join or blend to form a single entity"

**Home Assistant Source**
"a thing from which something originates"

Others I like less, but for suggestions sake and stimulating ideas.

Origins, Next, Up, Grid.

girkers:
What about Home Assistant Appliance (HAA)?

Isn't that what it is and by rights irrespective of the installation method shouldn't everything work the same anyway?

Home Assistant OS to me is a similar thing to HAA but when you say Appliance is you take it out of the box and it just works.

adr/0015-home-assistant-one.md
frenck: I agree with @amelchio on this actually. This is more about documenting than supporting I guess? At least there is a difference between those things. As we don't want the end up with tons of documentation for all hypervisors.

Kanga-Who:
> What about Home Assistant Appliance

I'm not sure lumping HA in the same category as a Toaster or Coffee Machine is a great idea. It's not an appliance, it's a piece of software.

girkers:
@Kanga-Who why not?

Isn't that what the ultimate goal is something that you can simply turn on and it works?

Yes I appreciate that it is is a piece of software but once installed works just like an appliance. Your DVR, your Smart TV, they are just devices with "pieces of software" I don't think there needs to be a delineation.

Kanga-Who:
> Isn't that what the ultimate goal is something that you can simply turn on and it works?
> 
> Yes I appreciate that it is is a piece of software but once installed works just like an appliance.

Except, HA doesn't behave in this way. Even a basic installation requires a reasonable amount of set up and configuration, and needs constant maintenance due to APIs breaking, updates, SD cards failing, etc.

It is not "set and forget" like you fridge in 99% of use cases.

girkers:
I get that and appreciate where you are coming from, but I thought that this was the idea of this thread was to have a set up, to work 99.9% of the time like your fridge?

If this is not the case I apologise for reading it wrong, I understood that once it is up and running and nothing changes shouldn't it just work?

If I cut my own bread and it is thicker than the toaster can handle that is an unhandled change, whereas if I buy store bought sliced bread all the time, it will always fit in the toaster.

thomasloven:
Appliance has several other meanings (e.g. https://en.wikipedia.org/wiki/Virtual_appliance)

Kanga-Who:
> Appliance has several other meanings (e.g. https://en.wikipedia.org/wiki/Virtual_appliance)

I am very well aware, however, the vast majority of people do not know or understand this meaning.  Look at it from a new users perspective and the understanding they will most likely have.

If this installation type is the suggested starting point, it will attract new users with limited skills and knowledge and the name could/should be representative of that.

jfmcarreira:
> I am very well aware, however, the vast majority of people do not know or understand this meaning. Look at it from a new users perspective and the understanding they will most likely have.

I agree with making it easier to install for new users. But the reason that new users don't know the meaning I don't think it is a reason for using it. I think we should go with something simple, meaningful and which is not used in some other meaning. This is the problem of using appliance or hub.

I agree with reserving the "hub" name for a piece of hardware which would come with HA.

girkers:
I appreciate that my suggestion of the name Appliance has generated discussion, however my meaning my be getting lost in translation. I may not be explaining myself very well. "Virtual Appliance" was never mentioned and is not the context of what I am meaning.

I would consider "Appliance" something that contains everything you need to get going "Out of the Box", you download, install and you are ready to go. Go back to my PVR example you take it out of the box, you plug it in and configure it, once it is done it will just work.

This is what I am proposing, you download a package that contains everything you need to install and get running; OS and software. Whether that be a Pi image, a Virtual Disk or otherwise, the end user experience will be the same.

If you look up the term "Appliance" on dictionary.com [](https://www.dictionary.com/browse/appliance?s=t) you will see that one of the definitions is "an instrument, apparatus, or device for a particular purpose or use."

And this is what I am suggesting is that once the installation and configuration is complete you will have an Home Assistant Appliance with nothing else to do but use it.

bdraco:
Home Assistant Hub 👍 
Home Assistant Complete
Home Assistant Experience

balloob:
I have addressed the comments and updated the name to Home Assistant All-in-One as that came out as the winner of [the polls](https://community.home-assistant.io/t/help-us-name-a-home-assistant-installation-method-polls-added/204166/159?u=balloob). 

Still have to decide on the Hypervisor. Currently adopted Pascal's list.

balloob:
I actually read the polls wrong, and Home Assistant OS has more votes. This is actually good because it would conflict with a previous installation method and AIO as abbreviation would not help.

I've updated the ADR accordingly.

balloob:
The OS will still be the OS. It's actually the same as the OS installs all the other parts. We will need to make sure to put a good issue creator picker in place to make sure issues end up in the right place.

PR #406, opened 2020-06-22
Small adjustment in Network Manager wording for ADR-0014

### LINKED ISSUE #403: ADR 14 - NetworkManager not to be actively used
## Context

ADR 14 states that

"NetworkManager is installed but not actively used."

Some add-ons (AdGuard for one) requires a fixed IP. As far as I am aware, the only way to achieve that is via NetworkManager.

It is also at odds with this documentation.

https://github.com/home-assistant/operating-system/blob/dev/Documentation/network.md#static-ip

## Proposal

The restriction is modified to be that "NetworkManager is installed but not actively used other than to set a Fixed IP & Gateway and Static DNS."

## Consequences

Honestly, I don't know, I just know that the install instructions for the add-on are now contradicted by the ADR.

frenck:
- Small wording change on the use of the NetworkManager, fixes #403
- Adjusted list char used to make it consistent with the rest of the document

### DIFF
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -39,7 +39,7 @@ This installation method can easily be broken if one manages the operating syste
 - All system dependencies are installed according to the manual.
 - No additional software, outside of the Home Assistant ecosystem, is installed.
 - Docker needs to be correctly configured to use overlayfs2 storage and journald as the logging driver.
-- NetworkManager is installed but doesn't need to be actively used.
+- NetworkManager is installed and enabled in systemd, but doesn't need to be actively used.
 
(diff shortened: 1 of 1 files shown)


pvizeli:
That's wrong. NetworkManager is force required and need active used for an support system. I'm fine to change that if someone will add systemd networkd support too

frenck:
Yeah @pvizeli, that is what I tried to re-word it too. I see the confusion caused by this one as well.
Adjusting.

frenck:
@pvizeli Made a small tweak to make it more clear? Is this acceptable?

pvizeli:
Yeah I think that is fine :)

PR #407, opened 2020-07-02
ADR-16: Home Assistant Core installation method

balloob:
And finally, the last one :)

Initially I was planning on redirecting users that want to try this installation method to the developer documentation. However, the dev docs talk about cloning the repository versus HA Core which is installed from PyPI. Therefore I have removed that clause.

### DIFF
--- a/adr/0016-home-assistant-core.md
+++ b/adr/0016-home-assistant-core.md
@@ -18,7 +18,7 @@ This is for running just the Home Assistant Core application directly on Python.
 
 - All major Linux distributions, latest stable and major versions.
 - Windows; only using WSL.
-- macOS; Python via Brew.
+- macOS; Python via Homebrew.
 
 ### Supported Python versions
 
--- a/adr/0016-home-assistant-core.md
+++ b/adr/0016-home-assistant-core.md
@@ -1,0 +1,63 @@
+# 0016. Installation method: Home Assistant Core
+
+Date: 2020-07-01
+
+## Status
+
+Accepted
+
+## Context
+
+Define a supported installation method as per [ADR-0012](https://github.com/home-assistant/architecture/blob/master/adr/0012-define-supported-installation-method.md).
+
+## Decision
+
+This is for running just the Home Assistant Core application directly on Python. It does not provide the full Supervisor experience and thus does not provide the Supervisor panel and add-ons.
+
+### Supported Operating Systems and versions
+
+- All major Linux distributions, latest stable and major versions.
+- Windows; only using WSL.
+- macOS; Python via Brew.
+
+### Supported Python versions
+
+Running Home Assistant Core is only supported when running the application using the official Python virtual environment. Running Home Assistant Core without a virtual environment, system/globally installed Python packages, is not supported.
+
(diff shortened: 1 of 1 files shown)


CoMPaTech:
Just as a thought, like the 'security patching of the os' maybe exclude handling of AppArmor/SElinux, ufw/firewalld and the like? So you'll heads-up the potential user on that as well?

adr/0016-home-assistant-core.md
```diff
@@ -0,0 +1,63 @@
...
+  Requires installing Python 3 with venv support (the default except on Debian based systems). Then create a virtual environment and install Home Assistant Core via pip.
+
+  For packages that require compilation, the user will need to install compilers and other development packages. If those development packages are not the same as provided by the operating system, you can break your system.
+
+* **Start when the system is started:** This is the responsibility of the user. It is based on their operating system.
+* **Run with full network access:** Works, is the only option.
+* **Access USB devices:** This works out of the box.
+
+* **Maintaining the Home Assistant installation**
+  Maintenance can be hard.
```
DubhAd, line 41: > Maintenance requires more time, effort, skills, and experience than the other methods.

balloob:
```suggestion
  Maintenance requires more time, effort, skills, and experience than the other methods.
```

adr/0016-home-assistant-core.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Decision
+
+This is for running just the Home Assistant Core application directly on Python. It does not provide the full Supervisor experience and thus does not provide the Supervisor panel and add-ons.
+
+### Supported Operating Systems and versions
+
+- All major Linux distributions, latest stable and major versions.
+- Windows; only using WSL.
+- macOS; Python via Brew.
```
thomasloven, line 21: Homebrew

balloob:
```suggestion
- macOS; Python via Homebrew.
```

balloob:
@CoMPaTech I think that bring your own OS should be clear enough that the user is responsible for everything.

frenck:
I think we need to add a section on what is going to be documented. Obviously we need installation guides (which is in the current ADR).

But do we want to impose a limit on things like documenting dependencies and such?

As for our current documentation, it is riddled with exceptions on additional packages/dependencies to install and a-like. Considering the experience level set, is that really needed? Imho it confuses a lot in the docs.

DubhAd:
> I think we need to add a section on what is going to be documented. Obviously we need installation guides (which is in the current ADR).
> 
> But do we want to impose a limit on things like documenting dependencies and such?

I think we should document depedencies, but it should be clearly tagged as only applying to the Core install, and probably at the end of the page. I don't think it _needs_ to be detailed, though it would help. As a minimum, for `stream` for example:

> ### Core install dependencies
> 
> Users of the core install method will need to install FFmpeg 4.0 or later. Other install methods include all required dependencies already.

or for the `bluetooth_tracker`:

> ### Core install dependencies
> 
> Users of the core install method will need to install Bluetooth development libraries for their operating system, to enable `pybluez` to be installed. Other install methods include all required dependencies already.

That provides enough information that the user can identify what needs to be installed

balloob:
Okidoki, what about this @DubhAd & @frenck ?

PR #422, opened 2020-08-14
Mention existing integrations in ADR0007

MartinHjelmare:
- Add language about changing config YAML in a platform section for an existing integration:
https://github.com/home-assistant/architecture/pull/361#issuecomment-609826815

### DIFF
--- a/adr/0007-integration-config-yaml-structure.md
+++ b/adr/0007-integration-config-yaml-structure.md
@@ -33,6 +33,7 @@ C. Harder for Home Assistant to support confused users and contributors and hard
 ## Decision
 
 We limit the configuration YAML structure to one way for new integrations. We require all YAML configuration, if present, for an integration to be located under the integration domain key in configuration YAML, for all new integrations (point 2 above).
+
 For existing integrations we don't allow changes to YAML configuration in platform sections until the integration has been refactored and the configuration moved under the integration domain key.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


thomasloven:
May discourage bugfixes, but I assume exceptions may still be granted on a case-by-case basis?

adr/0007-integration-config-yaml-structure.md
```diff
@@ -33,6 +33,7 @@ C. Harder for Home Assistant to support confused users and contributors and hard
 ## Decision
 
 We limit the configuration YAML structure to one way for new integrations. We require all YAML configuration, if present, for an integration to be located under the integration domain key in configuration YAML, for all new integrations (point 2 above).
+For existing integrations we don't allow changes to YAML configuration in platform sections until the integration has been refactored and the configuration moved under the integration domain key.
```
balloob, line 36: If you want this on a new line, either add one before, or add 2 spaces to end of previous line. Now it's rendered as 1 paragraph.

balloob:
Bug fixes mostly don't need YAML config changes.

thomasloven:
The key word is "mostly". API changes may, for example.
But let's cross that bridge when we get there.

adr/0007-integration-config-yaml-structure.md
MartinHjelmare: I didn't intend to add a paragraph. Should we add one?

MartinHjelmare:
If the community wants to maintain an integration it will need to be updated to follow our current design sooner or later. Here we draw the line that changing configuration YAML requires an up to date integration structure.

Issue #427, opened 2020-08-19
Supported Operating System, System dependencies and versions: NetworkManager >= 1.18.0 ?


danielbrunt57:
network-manager in Debian 10 buster (stable) is 1.14.6.  
network-manager 1.18.0 is listed as experimental (net): 1.18.0-1 [debports]: powerpcspe. 
Should I be using 1.18.0-1 for a supported supervised installation on Debian 10?

danielbrunt57:
I updated docker to use journald instead of the default json-file as the logging driver and now Supervisor no longer warns me that I'm using an unsupported platform so network-manager 1.14.6 must be okay.

carstencodes:
For me, the workaround mentioned by @danielbrunt57 has not worked. Searching for [network-manager on debian stable](https://packages.debian.org/search?suite=all&searchon=names&keywords=network-manager) resolves 1.14.6-2. Therefore debian testing (1.26.2-1) or debian unstable (same version) must be used, which conflicts with debian stable (buster). Can you please clarify this issue?

danielbrunt57:
> 
> 
> For me, the workaround mentioned by @danielbrunt57 has not worked. Searching for [network-manager on debian stable](https://packages.debian.org/search?suite=all&searchon=names&keywords=network-manager) resolves 1.14.6-2. Therefore debian testing (1.26.2-1) or debian unstable (same version) must be used, which conflicts with debian stable (buster). Can you please clarify this issue?

Your supervisor warns you about an unsupported platform despite using Network Manager 1.14.6-2 and journald for docker logging?  Any errors or warnings in your supervisor log?

carstencodes:
Ok, I see.

20-08-25 19:02:39 ERROR (MainThread) [supervisor.core] Detected unsupported OS: Raspbian GNU/Linux 10 (buster)
20-08-25 19:02:40 INFO (MainThread) [__main__] Run Supervisor
20-08-25 19:02:40 INFO (MainThread) [supervisor.api] Start API on 172.30.32.2
20-08-25 19:02:40 CRITICAL (MainThread) [supervisor.core] System running in a unsupported environment!

Do I really need to install Debian/arm 10 to make the supervisor happy?

danielbrunt57:
[...]

PR #431, opened 2020-09-04
Adjust required version of NetworkManager

cerebrate:
Changes the required version of NetworkManager to match that available in Debian 10 (buster), since 1.18.0 is not. Since supervisor reports no errors running with 1.14.6, this should be a harmless change, and eliminates confusion in diagnosing such errors, as seen here - https://community.home-assistant.io/t/networkmanager-1-18-0-on-debian-buster/223968 - and in issue #427 .

### DIFF
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -24,7 +24,7 @@ Docker CE (Community Edition) is the only supported containerization method for 
 
 - Docker CE >= 19.03
 - Systemd >= 239
-- NetworkManager >= 1.18.0
+- NetworkManager >= 1.14.6
 - Avahi >= 0.7
 - AppArmor == 2.13.x (built into the kernel)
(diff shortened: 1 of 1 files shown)


frenck:
This is not entirely true/correct. Originally, the Network manager was bumped to support WireGuard (networkmanager 1.16+). So, this might have future consequences. I do realize this feature of the network manager is currently not used though).

cerebrate:
Hm. I do see the problem there, but at the moment, the seeming need for an impossible configuration (at least without mixing distro versions or compiling your own from source) seems like a bigger issue? At least until/unless a later NetworkManager makes it into buster-backports.

frenck:
I think it is fine to merge like this.

@pvizeli ?

frenck (approved):

ludeeus:
Version 1.14.6 does not work with our current implementation

pvizeli (approved):

Issue #270, opened 2019-08-06
Webscraping for local devices


amelchio:
## Context

In [ADR-0004](https://github.com/home-assistant/architecture/blob/master/adr/0004-webscraping.md) it was decided that webscraping will not be allowed, with webscraping being defined like this:

> Webscraping is when we use code to mimic a user and log in to a website and get data in Home Assistant.

However, that definition has been found to be quite ambiguous:

1. It does not distinguish between data formats. We have actually now banned structured data formats like JSON which was apparently not the intention.
2. The definition seems to only apply to websites with authentication.
3. It is unclear what "a website" means, in particular whether local appliances with an embedded webserver (such as routers and printers) are included.

The first two points just look like oversights (or maybe even nitpicking) but the third point has caused real confusion. This new issue is to continue the decision process since the original issue #252 has already been closed.

## Proposal

* We disallow integrations that get specific values from publicly hosted HTML pages.
* We do allow integrations that read structured data like JSON or XML.
* We do allow integrations that interpret HTML from devices on the local network. This can use e.g. regular expressions and DOM parsing but not a full-blown browser engine (even if headless).

## Consequences

ADR-0004 becomes less ambiguous. Unauthenticated access is included in the ban. Data exchange formats are excluded from the ban. Local devices are explicitly exempted from the ban.

amelchio:
So my proposal is to allow scraping of local devices. I think local devices generally do not have the downsides that the original issue wanted to counter. 

Firmware is rarely updated significantly once a device has shipped so the HTML structure does not change like a public website does. When it does, fixing it is easier than fixing a reverse engineered binary protocol.

Banning such integrations seems to throw the baby out with the bathwater.

frenck:
I don't see the issue with the "banning". Since clarification is important: Banning is an inappropriate word to use. 

We don't block them, we currently do not allow them to be part of the core repository. However, you can still use and distribute them as part of custom integration.

pvizeli:
> Firmware is rarely updated significantly once a device has shipped so the HTML structure does not change like a public website does

What should we do if 50% of our user have the new firmware and other 50% not? I guess the library need handle that and not HA but that will be dificult. With custom component, user can run what every they need to parse here website.

ties:
It comes down to balancing user value and maintenance. Compared to other domotics product, home assistant tends to go for user value.

Using the line of reasoning by @pvizeli and @Swamp-Ig, proposed text:
[...]

PR #512, opened 2021-02-09
Amend ADR004 to allow extracting forms during the auth phase

balloob:
We've seen more and more integrations struggle with ADR004. The problem they struggle with is not that the data fetching is based on scraping, but that some parts of the authentication flow requires extraction of the form fields and their values. Fetching data should happen via an API. We will be more flexible, and hinder less integrations, if we allow integrations to extract forms during the authentication flow. 

That's why I want to propose that we amend ADR004 to allow web scraping of forms during the authentication phase. For robustness, integrations should fetch all fields of the form and not individually extract values of fields in the form. These are less likely to change.

Please limit discussion in this PR to this change. Discussing the ADR004 or other amendments is out of scope for this PR. Start a discussion topic instead.

### DIFF
--- a/adr/0004-webscraping.md
+++ b/adr/0004-webscraping.md
@@ -1,5 +1,4 @@
 # 4. Webscraping
-gs
 Date: 2019-06-27
 
 Architecture issue: [#252](https://github.com/home-assistant/architecture/issues/252)
--- a/adr/0004-webscraping.md
+++ b/adr/0004-webscraping.md
@@ -4,6 +4,10 @@ Date: 2019-06-27
 
 Architecture issue: [#252](https://github.com/home-assistant/architecture/issues/252)
 
+Changelog:
+ - 2019-06-27 Initial version accepted
+ - 2021-02-09 Amendment to add exception for authentication phase
+
 ## Status
 
 Accepted
--- a/adr/0004-webscraping.md
+++ b/adr/0004-webscraping.md
@@ -27,9 +27,10 @@ Webscraping comes with the following downsides:
 - We no longer accept any new integration that relies on webscraping
 - We identify, deprecate for 2 releases and remove integrations that rely on webscraping
 - It will still be possible to have custom integrations provide information via webscraping
-- Generic integration to parse HTML are excluded from this decision
 
-### Exception
+### Exceptions
+
+Generic integrations to parse HTML are excluded from this decision.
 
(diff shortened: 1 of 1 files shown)


adr/0004-webscraping.md
```diff
@@ -1,5 +1,5 @@
 # 4. Webscraping
-
+gs
```
MartinHjelmare, line 2:
```suggestion
```

frenck, line 2: ?

balloob: haha oops that's me thinking I'm in my terminal 😬

frenck:
I think this is fine. Should we add an update date or changelog or something to these?

balloob:
Added a changelog.

frenck (approved):

pvizeli (approved):

adr/0004-webscraping.md
```diff
@@ -25,6 +29,10 @@ Webscraping comes with the following downsides:
 - It will still be possible to have custom integrations provide information via webscraping
 - Generic integration to parse HTML are excluded from this decision
 
+### Exception
```
ludeeus, line 32: We already have an exception "Generic integration to parse HTML are excluded from this decision"
Should not the new "Exception" section should include both?

ludeeus (approved):

bramkragten (approved):

bdraco (approved):

bdraco (approved):

raman325 (approved):

rohankapoorcom (approved):

KimGuang44 (approved):
感恩您，煩請下次更改時附註一封中文信件說明修改了什麼，謝謝

alandtse (approved):

amelchio:
I am fine with the change. However, an ADR should not just document _what_ has been decided but also _why_.

In #270 webscraping of local devices was rejected. Top arguments included webscraping being a horror that must unconditionally be avoided and the observation that a local website is still a webstite. This document should make it clear to future contributors (and maybe me) why those concerns do not apply to login pages.

The reasoning in the description does not seem to explain it, it sounds more like a contradiction to me:

> The problem is not that the API is based on scraping, but that some parts of the login form require extraction of all the form fields.

balloob:
I've updated the initial comment.

PR #514, opened 2021-02-11
Fix broken link to dev docs.

MatthewFlamm:
Current link points to outdated developers doc location.

I pointed to the subsection specifically about `entity_registry_enabled_default`, but we could also consider just pointing to the whole page.  It would be less specific to point to the whole page, but more robust to changes in the structure within the page.


frenck (approved):
Thanks, @MatthewFlamm 👍

Issue #496, opened 2021-01-31
Please use discussions instead of issues for new architectural proposals! Thanks!


frenck:
If you want to propose an architectural change, please use the new Discussion instead of creating a new issue.

https://github.com/home-assistant/architecture/discussions

Thanks! 👍

PR #521, opened 2021-02-23
Change guidance from issue to discussion in README

iMicknl:
Not sure if `Discussions` is here to stay, but for now it would be good to change this link in the readme. Since the link is directly to the new issue page, you won't see the pinned issue (https://github.com/home-assistant/architecture/issues/496).


frenck (approved):
Thanks, @iMicknl 👍

PR #543, opened 2021-04-08
Add clearer deps instructions

sveip:
..on which dependencies are needed and how to install them.

### DIFF
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -1,6 +1,6 @@
 # 0014. Installation method: Home Assistant Supervised
 
-Date: 2020-06-03
+Date: 2021-05-08
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


adr/0014-home-assistant-supervised.md
```diff
@@ -20,14 +20,25 @@ This method is considered advanced and should only be used if one is an expert i
 
 ### Supported Operating System, System dependencies and versions
 
-Docker CE (Community Edition) is the only supported containerization method for Home Assistant Supervised. We only support FHS 3.0 on the host file system.
+Docker CE (Community Edition) is the only supported containerization method for Home Assistant Supervised. It can be installed in many different ways, one of them is using the script here https://get.docker.com/ another is adding the repo manually like described here https://www.linode.com/docs/guides/how-to-install-docker-ce-on-debian-10/. We only support FHS 3.0 on the host file system.
```
frenck, line 23: This is not a guide providing installation instructions, but this is an architectural decision record. This does not belong here.

adr/0014-home-assistant-supervised.md
```diff
@@ -20,14 +20,25 @@ This method is considered advanced and should only be used if one is an expert i
...
+The installer script will check for the following pre-required packages:
+
+- systemctl
+- nmcli
+- apparmor_parser
+- docker
+- jq
+- curl
+- dbus-daemon
+
```
frenck, line 41: This is a duplication of what is listed above already?

frenck (changes requested):
This is not a guide for providing installation instructions.

adr/0014-home-assistant-supervised.md
sveip: Not really, it's nowehere documented that e.g. `jq` is required, and it's not part of the standard installer of Debian. So if you would follow the req's and install `docker-ce` you will get an error that it's missing. As you say, maybe it should be up to the user to figure out?

adr/0014-home-assistant-supervised.md
sveip: I agree when you put it like that, but it's linked to here as "requirements", https://github.com/home-assistant/supervised-installer, so maybe that page needs re-wording?

As an example, atm the insaller checks for `docker`, and will pass even if the user installed the apt upstream non community edition. I just thought it would make it a bit clearer to explain what is checked against with regards to deps.

adr/0014-home-assistant-supervised.md
frenck: Yes, if you need this level of guidance, you should not be running this installation method. We strongly recommend using the Home Assistant Operating System or Home Assistant Container installation methods instead.

As said, this is not an installation guide.

adr/0014-home-assistant-supervised.md
frenck: It is linked to the requirements that are listed below.

Your text adds third-party installation instructions, which is not what this document is for.

adr/0014-home-assistant-supervised.md
sveip: I totally understand, sorry for wasting your time, it's highly appreciated. The add-on ecosystem is very nice, so for hw that isn't a nuc, odroid or an pi this method is still very valuable. :+1:

PR #612, opened 2021-08-21
Update Supervised supported OS

frenck:
Updates the ADR that lists the requirements for Home Assistant Supervised installations.

Marks Debian 11 supported, removes Debian 10 as it is deprecated to keep documentation current and reflect what users should be running.

ℹ️  This PR correctly targets the `current` branch, as the Supervisor is released independently of Home Assistant Core.
⚠️ Should not be merged until the related Supervisor is released to stable

After the deprecation period for Debian 10 has been passed, we can look at bumping other dependencies to have a higher minimum requirement.

Supervisor PR: https://github.com/home-assistant/supervisor/pull/3066

### DIFF
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -26,7 +26,7 @@ Docker CE (Community Edition) is the only supported containerization method for 
 - Systemd >= 239
 - NetworkManager >= 1.14.6
 - AppArmor == 2.13.x (built into the kernel)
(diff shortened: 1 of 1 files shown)


ludeeus (commented):
We should remove "10" from lines 49 and 68, so they only state "Debian" or update those as well

frenck:
Done, removed it, making it easier for future updates.

ludeeus (approved):

pvizeli (approved):

PR #644, opened 2021-10-20
ADR: Supported database engines

emontnemery:


adr/0018-supported-database-engines.md
```diff
@@ -0,0 +1,53 @@
...
+
+Date: 2021-10-20
+
+## Status
+
+TBD
+
+## Context
+
+We currently have no restrictions on supported database engines for the recorder, and the documentation states:
```
frenck, line 11:
```suggestion
We currently have no restrictions on supported databases for the recorder, and the documentation states:
```

adr/0018-supported-database-engines.md
```diff
@@ -0,0 +1,53 @@
...
+
+ > Home Assistant uses SQLAlchemy, which is an Object Relational Mapper (ORM). This means that you can use any 
+ > SQL backend for the recorder that is supported by SQLAlchemy, like MySQL, MariaDB, PostgreSQL, or MS SQL Server.
+
+This claim is not correct, SQLAlchemy will behave differently on different database engines, and features relied
+on by the recorder may work differently, or not at all, in different database engines.
```
frenck, line 17:
```suggestion
This claim is not correct, SQLAlchemy will behave differently on different databases, and features relied
on by the recorder may work differently, or not at all, in different databases. 
```

adr/0018-supported-database-engines.md
```diff
@@ -0,0 +1,53 @@
...
+
+As a matter of fact, we have a lot of special handling for the different databases out there, which makes
+maintenance and implementing changes really hard. For contributors, it is impossible to test changes against
+all database services SQLAlchemey could work with. This often causes bugs being reported for the less common
+databases after a new release.
+
+TL;DR: We advertise a lot of different databases being supported, without version limitation mentioned. This
+causes the end-user to think their database and version of choice will work with Home Assistant. Unfortunately,
+that is not always the case; The user experience for those using a less popular database engine or an outdated
+version is very poor, with frequent issues preventing the recorder from working after an upgrade of Home Assistant.
```
frenck, line 27:
```suggestion
that is not always the case; The user experience for those using a less popular database or an outdated version
is very poor, with frequent issues preventing the recorder from working after an upgrade of Home Assistant.
```

adr/0018-supported-database-engines.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Decision
+
+Support a limited set of database engines, and also limit the supported versions for the supported engines
+in a similar way as we have limited installation types for Home Assistant. This greatly reduces the
```
frenck, line 32:
```suggestion
Support a limited set of databases, and also limit the supported versions for the supported databases
in a similar way as we have limited installation types for Home Assistant. This greatly reduces the
```

adr/0018-supported-database-engines.md
```diff
@@ -0,0 +1,53 @@
...
+to start with Home Assistant 2022.2.
+
+The above listed minimal versions may change over time. In case a minimal version bump is required, this will
+be announced as a breaking change, with a depreciation period of 2 release cycles (2 months).
+
+
+## Consequences
+
+Existing users of database engines that are not in the supported list above, will have to move to a different
+database solution. The recommendation is to use the built-in SQLite database when possible.
```
frenck, line 53:
```suggestion
Existing users of databases that are not in the supported list above, will have to move to a different
database solution. The recommendation is to use the built-in SQLite database when possible.
```

adr/0018-supported-databases.md
```diff
@@ -0,0 +1,53 @@
...
+to start with Home Assistant 2022.2.
+
+The above listed minimal versions may change over time. In case a minimal version bump is required, this will
+be announced as a breaking change, with a depreciation period of 2 release cycles (2 months).
+
+
+## Consequences
+
+Existing users of databases that are not in the supported list above, will have to move to a different
+database solution. The recommendation is to use the built-in SQLite database when possible.
```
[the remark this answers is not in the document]
rohankapoorcom: That's my point though.

It's the default, which means users don't switch away from it unless they have a reason to (moving database to a separate machine, database size, etc). Those are the users that will be affected by their chosen database no longer being supported. For them, switching back to SQLite would not make sense, upgrading to the latest supported version of their chosen Database would.

I agree the general recommendation should stay SQLite, but that's not how I read that line.

adr/0018-supported-databases.md
```diff
@@ -0,0 +1,53 @@
...
+is very poor, with frequent issues preventing the recorder from working after an upgrade of Home Assistant.
+
+## Decision
+
+Support a limited set of databases, and also limit the supported versions for the supported databases
+in a similar way as we have limited installation types for Home Assistant. This greatly reduces the
+maintenance effort, making it easier to implement and test new features and giving the end-user a better
+experience with reliable upgrades.
+
+Limit DB-support to the following:
```
spacegaier, line 36:
```suggestion
Limit DB support to the following:
```
I dunno, but this hyphen looks weird / out of place. Probably me being too nit-picky...

PR #676, opened 2021-12-03
Remove broken link in translations adr

eikowagenknecht:
Link is not working any more because hastebin has been aquired by toptal.com.

I think it is okay to just remove it as it is not neccessary to understand the ADR.

If you think it *is* needed, an example of the old API call would need to be hosted on a more reliable url (maybe in this repo?).


frenck (approved):
Thanks, @eikowagenknecht 👍

PR #693, opened 2021-12-20
0019 GPIO: Spelling consistency plus final approval date correction

spacegaier:
There were two inconsistent spellings of "I2C". Plus shouldn't the date be today as this ADR was approved and merged today?


frenck (approved):
Thanks, @spacegaier 👍

PR #718, opened 2022-01-27
Add exception for unique_id to ADR0007

ludeeus:
Ref: <https://github.com/home-assistant/core/pull/58596#issuecomment-1004117319>

### DIFF
--- a/adr/0007-integration-config-yaml-structure.md
+++ b/adr/0007-integration-config-yaml-structure.md
@@ -49,7 +49,7 @@ sensor:
 
 ### Exceptions
 
-- Implementing an optional `unique_id` configuration key to set the unique ID of the entity created by the platform is allowed.
+- For integrations that do not fall under [ADR-0010](https://github.com/home-assistant/architecture/blob/master/adr/0010-integration-configuration.md), implementing an optional `unique_id` configuration key to set the unique ID of the entity created by the platform is allowed.
 
 ## Consequences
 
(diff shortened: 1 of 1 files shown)


frenck (approved):

adr/0007-integration-config-yaml-structure.md
```diff
@@ -47,6 +47,10 @@ sensor:
     username: user
 ```
 
+### Exceptions
+
+- Implementing an optional `unique_id` configuration key to set the unique ID of the entity created by the platform is allowed.
```
MartinHjelmare, line 52: Should we say something about what platforms are allowed to do this. It's not allowed for all platforms, only platforms that create entities that derive their state from other integrations or platforms for transport implementing integrations, or if we have made some special exception.

frenck: Maybe say: May not conflict with ADR-0010?

ludeeus: It has been added to `command_line` that is not excepted from ADR0010 or derived from other entities or is a transport protocol, so changing the wording here does not reflect what has already been accepted.

frenck: The command line defines its own "protocol". As in, it does not connect to a device or service in a normal sense. It is safe to say it doesn't fall under adr-0010 (at least not intentional)

ludeeus, line 52:
```suggestion
- For integrations that do not fall under [ADR-0010](https://github.com/home-assistant/architecture/blob/master/adr/0010-integration-configuration.md), implementing an optional `unique_id` configuration key to set the unique ID of the entity created by the platform is allowed.
```

MartinHjelmare (approved):

frenck (approved):

PR #721, opened 2022-01-30
Create blank.yml

moovy2:


frenck:
We don't need a blank workflow :)

PR #749, opened 2022-03-06
Loosen requirement in ADR10 how UI-based configuration is stored

balloob:
This is a small update to ADR010 to remove the implementation detail of how UI-based configuration is stored in the backend (previously: config flow, storage collection).


elupus:
Looks good to me. Do we want to require backward yaml compat?

MartinHjelmare (approved):

elupus (approved):

balloob:
Nah, compatibility is not a goal. We just continue to support both as we do today.

PR #767, opened 2022-05-10
Remove OS version

ludeeus:


adr/0015-home-assistant-os.md
```diff
@@ -31,7 +31,7 @@ This is the generally recommended installation method and the one that our websi
 
 ### Supported Operating Systems and versions
 
-This installation method is only available and supported using the Home Assistant Operating System. Only the latest major version is supported (Currently, 6.x).
+This installation method is only available and supported using the Home Assistant Operating System. Only the latest major version is supported (Currently, 8.x).
```
balloob, line 34: Let's just drop the current version. 
```suggestion
This installation method is only available and supported using the Home Assistant Operating System. Only the latest major version is supported.
```

frenck (approved):

PR #801, opened 2022-09-09
Supervised journal gateway / udisk / docker

pvizeli:
https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/rootfs-overlay/usr/lib/systemd/system/systemd-journal-gatewayd.socket
https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/rootfs-overlay/etc/docker/daemon.json

### DIFF
--- a/adr/0014-home-assistant-supervised.md
+++ b/adr/0014-home-assistant-supervised.md
@@ -22,9 +22,10 @@ This method is considered advanced and should only be used if one is an expert i
 
 Docker CE (Community Edition) is the only supported containerization method for Home Assistant Supervised. We only support FHS 3.0 on the host file system.
 
-- Docker CE >= 20.10
+- Docker CE >= 20.10.17
 - Systemd >= 239
 - NetworkManager >= 1.14.6
+- udisks2 >= 2.8
 - AppArmor == 2.13.x (built into the kernel)
 - Debian Linux Debian 11 aka Bullseye (no derivatives)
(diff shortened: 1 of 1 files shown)


adr/0014-home-assistant-supervised.md
```diff
@@ -38,8 +39,9 @@ This installation method can easily be broken if one manages the operating syste
 - The operating system is dedicated to running Home Assistant Supervised.
 - All system dependencies are installed according to the manual.
 - No additional software, outside of the Home Assistant ecosystem, is installed.
-- Docker needs to be correctly configured to use overlayfs2 storage, journald as the logging driver, and cgroup v1.
+- Docker needs to be correctly configured to use overlayfs2 storage, journald as the logging driver with Container name as Tag, and cgroup v1.
```
mdegat01, line 42: >  with Container name as Tag

It needs to be in a specific tag right? Otherwise we won't be able to look up logs for particular addons/core/etc. Are we going with `CONTAINER_NAME` or did we want the name as the syslog identifier.

mdegat01 (changes requested):

adr/0014-home-assistant-supervised.md
pvizeli: I don't know, at least we should make sure people run it as same as on OS

mdegat01: Oh I see, you don't specify the tag name. We're just telling them to do this:
https://github.com/home-assistant/operating-system/blob/a6445af71282045c2ea92f1dea8d2f9e518bd008/buildroot-external/rootfs-overlay/etc/docker/daemon.json#L6-L8

That works fine then

mdegat01 (approved):

PR #829, opened 2022-11-03
Include `systemd-journal-remote`

mdegat01:


adr/0014-home-assistant-supervised.md
```diff
@@ -26,6 +26,7 @@ Docker CE (Community Edition) is the only supported containerization method for
 - Systemd >= 239
 - NetworkManager >= 1.14.6
 - udisks2 >= 2.8
+- systemd-journal-remote >= 239
```
pvizeli, line 29: Distro split it up, but in the end it's part of systemd and there build process. I don't think it have to be listed separate from requirements. We also not list all packages, i.e. just AppArmor and people have to deal at his own which sub packages it is needed.

agners: Yes, I agree 💯

ThomDietrich: Hence why I was surprised to see no rationale for this PR from @mdegat01 🙃

mdegat01: Sorry I had been chatting with Pascal and was reacting to all the discussion around this in the release notes. I thought we were supposed to list all software the user needed to install and this was an oversight when we added the journal gateway service. I'm fine if we don't think it's needed though.

frenck: The release has been gone for a bit now,  and the general storm seems to have lied down. Documentation has been adjusted where needed, and the more-info page provides help in case everything else was missed.

So, let's close this for now.

PR #914, opened 2023-05-30
Supported Python versions tweaks

frenck:
This PR addresses language comments by @MartinHjelmare to ADR-0020.

Additionally, it adjusts ADR-0016 to point to ADR-0020 (instead of previously copying ADR-0002, which is now defunct); as commented by @ludeeus on Discord.


MartinHjelmare (approved):
Thanks!

PR #990, opened 2023-10-19
Update supported boards

agners:
Add Home Assistant Yellow/Green to supported boards list. Also add ODROID-C4 and ODROID-M1.

Mention generic x86-64 support.

Also clarify virtualization support a bit and sort lists alphabetically.


frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #1002, opened 2023-11-21
Update minimum supported SQLite version to 3.40.1

bdraco:
Since we do not test against older versions of SQLite, they are a terrible user experience, and we keep [finding problems](https://github.com/home-assistant/core/pull/104296)

Update the minimum SQLite version to 3.40.1 which is the bookworm version

core PR proposes 2025.2 as the removal date for the old version https://github.com/home-assistant/core/pull/104298


Jc2k:
Sounds like a good idea to me 👍

jbouwh (approved):
LGTM,
Thnx @bdraco 👍

PR #1189, opened 2025-01-22
Update ADR-0015: sync list of supported targets with reality

sairon:
Updating the outdated list of targets to current state which the support stabilized at without any prior architecture discussion.

Commenting on individual changes:

 - On Yellow we rolled out official support for CM5 in HAOS 14. There was no architecture discussion on this and we should keep it in mind for the future. This is just reflecting the current state.
 - Similar for RPi 5 - we started to support it without any change here.
 - "Tinkerboard" was too vague, we don't support all models, so clarifying on the actual models currently supported.
 - N2+ is just a variation of N2, N2 is discontinued but can be interchanged easily with only a difference in the device tree. Explicitly adding also the N2+ to avoid confusion.
 - ODROID-M1S is officially supported for a while through a community addition. Unlike N2 vs. N2+, it's entirely different HW, so it deserves an extra line.
 - For x86, we need to state that not all hardware combinations can be fully supported - although the status quo is that the support is done on "best effort basis", we need to communicate it publicly too.
 - There was no mention of generic-aarch64 image. Although it can be targeted at many different boards, OS maintainers can't guarantee official support on any hardware. That's why the mention was added only to the VM line.

I am intentionally ignoring the presence of RPi 2 target in our builds. Support for that was probably never communicated publicly, this document neither the docs mention it and it's way too underspec'd for the current demands. These facts will eventually make it easier to drop it from our builds.

### DIFF
--- a/adr/0015-home-assistant-os.md
+++ b/adr/0015-home-assistant-os.md
@@ -18,20 +18,23 @@ This is the generally recommended installation method and the one that our websi
 
 ### Supported boards/hardware/machines
 
-- Home Assistant Yellow
+- Home Assistant Yellow (with CM4 and CM5 modules)
 - Home Assistant Green
 - Raspberry Pi 3 Model B and B+ 32-bit
 - Raspberry Pi 3 Model B and B+ 64-bit
 - Raspberry Pi 4 Model B 32-bit
 - Raspberry Pi 4 Model B 64-bit
-- Tinkerboard
+- Raspberry Pi 5 Model B 64-bit
+- ASUS Tinker Board
+- ASUS Tinker Board S
 - ODROID-C2
 - ODROID-C4
-- ODROID-N2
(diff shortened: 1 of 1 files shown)


frenck (changes requested):
We need to document the change and the date it changes in our ADRs.

sairon:
I see, but looking at the latest changes in the Git history, it's something that's been neglected in the past as well, so I need a bit more concrete specification how should it look. Is it enough if I add today's date as the last update date and the commit summary as a comment? The detailed summary of changes would be still available in Git.

frenck:
Sounds good to me, but let's at least make sure it is listed.

ADR-0004 seems to do it nicely

frenck (approved):
Thanks, @sairon 👍 

../Frenck

agners (commented):
LGTM!

PR #1285, opened 2025-09-29
Update ADR-0015: add already supported Khadas VIM3 to the list

sairon:
In #1189 Khadas VIM3 was unintentionally left out from the list of supported boards. Its support was added and maintained since home-assistant/operating-system#1473, so this is just another sync with the reality.


frenck (approved):
Thanks, @sairon 👍 

../Frenck

